Issue #1, opened 2017-01-02
Help wanted!


ivarconr:
Are you a go expert? We plan to implement a Unleash client for Go programming language. My go knowledge is about zero, and any help would be appropriated!

PR #2, opened 2017-03-06
Add a functional Go client.

jrbarron:
This branch should bring the Go client up to feature parity with the Node.js version which it is based on. It takes advantage of several Go features such as channels and Go routines for asynchronous communication. It is fully documented, but still lacking a complete set of autotests. I've started those, but there is still quite a lot of work before it has as much test coverage as the Node or Java versions.


sveisvei:
Great stuff. I do not yet have the insights into Go to spot bugs, but as far as I can tell this looks good. Appreciate that you are using the same "structure" as the Java and Node.js clients.

jrbarron:
@sveisvei Great! If you're not familiar with Go, there are a couple of features I can tell you about that might help you make sense of it:

- Functions can return multiple values and it is quite common to return an error as one of those. Ignoring errors in Go is considered very bad practice.

- Anything that starts with a capital letter is exported (ie: part of the public API), anything starting with a lower case letter is considered internal and can therefore not be accessed outside of this package. Agreeing on a good public API is perhaps the most important part of this merge request since obviously we don't want to remove anything that would break compatibility. All the public functions should be documented now. If you would like to see the docs, I can run a doc server locally and expose the port (via ngrok) for a short period of time. After the merge, I will publish on godoc.org (and setup travis CI)

- Any function called with `go` before it will start the function asynchronously in a separate "go routine" (lightweight thread). The easiest and safest way to communicate across threads is to use a Go concept called channels. Channels (like everything in Go) are strongly typed, but otherwise are very similar to web workers in javascript and passing messages. To write to a channel, we use this syntax:

```go
    myChan <- myInt
```

And to read, we use:

```go
    myInt := <- myChan
```

If a channel is unbuffered, only one thing can be written to a channel before it must be read by something else. Trying to write twice in a row to an unbuffered channel will block the second call until there is space in the channel. Alternatively, you can create a buffered channel of a specified size. This indicates how "slots" are available in that channel for writing before a read is required.

Finally the `:=` syntax that you see all over the place just creates a variable (using the type from the right hand side of the operator) and then assigns it. It's basically short hand notation.

```go
var myInt int
myInt = 42
// is the same as
myInt := 42
```

That should get you started! Anything else you are wondering about, just add a comment and I'll do my best to explain it!

ivarconr:
Overall a strong commit!! 

I'm not too familiar with go language, but there seems to be some issues with the `GradualRolloutUserId`-strategy implementation. Setting groupId=zero and generating userId's between (0, 100] gives true every-time, for all percentages (1-99). I suspect the normalize is not working as expected. 

If i change the groupId to "123" it gives true for all percentages above 50%!

jrbarron:
@ivarconr Ok, I'll take look and see if I can find out what's going on. I have some autotests for the normalize function so I'll start there.

ivarconr:
Hmm, the `gradualRolloutSessionId` seems to work as expected.

internal/strategies/gradualrolloutuserid.go
```diff
@@ -0,0 +1,44 @@
...
+		return false
+	}
+
+	groupIdValue := params[strategy.ParamGroupId]
+	groupId, ok := groupIdValue.(string)
+	if !ok {
+		return false
+	}
+
+	normalizedId := normalizedValue(ctx.SessionId, groupId)
```
ivarconr: SessionId?

jrbarron: *Ahem*, nope no copy-pasting there.. ;) I'll fix that.

client.go
```diff
@@ -0,0 +1,293 @@
...
+			return *opts.fallback
+		}
+		return false
+	}
+
+	if !f.Enabled {
+		return false
+	}
+
+	for _, s := range f.Strategies {
```
ivarconr: A feature toggle can have multiple strategies and one of them could be enabled, but not necessary the first one. Thus you have to continue if the first found strategy returns false until there is no more strategies.

jrbarron: Ah right. On it!

jrbarron:
@ivarconr@sveisvi I'm just looking at adding some autotests for the strategy implementations now. Go has a built-in testing framework, but it is quite low level in comparison to something like Ava. Alternatively I could add a dependency to something like Ginkgo (https://onsi.github.io/ginkgo/) which is BDD. Thoughts?

ivarconr:
I have no strong feeling as I am at "hello world" level in Go (but it is on my list of the language to invest in for 2017). 

My personal feeling is to have something that is easy to reason about, but I will go for whatever you are most comfortable with!

ivarconr:
I have manually tested all strategies and they all seems to work (except from the minor GradualRolloutUserId typo). Would be nice to have some test similar to the node-client tests where the gradual roll-out strategies are bench-marked to be within +/-1 percentage of the defined roll-out percentage.

jrbarron:
Alright I've fixed those two issues and added some autotests for the first strategy (the one that had the problems). I'll continue adding autotests for the other strategies over the next few days as I get time and then move on to the rest of the package.

internal/strategies/gradualrolloutuserid_test.go
```diff
@@ -0,0 +1,112 @@
+package strategies
+
+import (
+	"github.com/unleash/unleash-client-go/context"
+	"github.com/unleash/unleash-client-go/strategy"
+	"strconv"
+	"testing"
+)
+
+func TestGradualRolloutUserId_Name(t *testing.T) {
```
ivarconr: Nice, these are quite easy to read.  But how to I run test's in Go? 

```
➜  unleash-client-go git:(master) go test internal/strategies/gradualrolloutuserid_test.go
# command-line-arguments
internal/strategies/gradualrolloutuserid_test.go:11: undefined: NewGradualRolloutUserId
internal/strategies/gradualrolloutuserid_test.go:20: undefined: NewGradualRolloutUserId
internal/strategies/gradualrolloutuserid_test.go:22: t.Run undefined (type *testing.T has no field or method Run)
internal/strategies/gradualrolloutuserid_test.go:35: t.Run undefined (type *testing.T has no field or method Run)
internal/strategies/gradualrolloutuserid_test.go:48: t.Run undefined (type *testing.T has no field or method Run)
internal/strategies/gradualrolloutuserid_test.go:51: undefined: normalizedValue
internal/strategies/gradualrolloutuserid_test.go:65: t.Run undefined (type *testing.T has no field or method Run)
internal/strategies/gradualrolloutuserid_test.go:68: undefined: normalizedValue
internal/strategies/gradualrolloutuserid_test.go:82: t.Run undefined (type *testing.T has no field or method Run)
internal/strategies/gradualrolloutuserid_test.go:103: undefined: round
internal/strategies/gradualrolloutuserid_test.go:103: too many errors
FAIL	command-line-arguments [build failed]
```

internal/strategies/gradualrolloutuserid.go
```diff
@@ -0,0 +1,44 @@
+package strategies
```
ivarconr: What is the file-name conventions in go, snake-case, camel-case? Feels a bit uncommon to call a file `gradualrolloutuserid.go`.

internal/strategies/gradualrolloutuserid_test.go
jrbarron: @ivarconr For local paths, you need to start with `./` otherwise it assumes you are referring to a top level package in which you need the fully qualified path:

`go test github.com/unleash/unleash-client-go/internal/...`

internal/strategies/gradualrolloutuserid.go
jrbarron: @ivarconr This was more personal preference that any convention. I don't have strong feelings about it so we can change it to be more inline with the other repos. Files containing test cases that should not be compiled into the binary have an `_test` suffix so we can probably use underscores between words in filenames? That pattern seems to be used in other Go projects at least and then we avoid mixing hyphens and underscores. Agreed?

ivarconr: sounds reasonable!

ivarconr:
Let me know when you feel this is ready to merge!

SimenB:
Close #1? :smile:

Issue #3, opened 2017-06-07
Make Unleash Globally accessible


shivid:
Unleash takes some time to initialize before I can actually run Unleash.IsEnabled, but I don't have a globally accessible method to do that. Right now my only option is to make a NewClient for every package and access the IsReady channel that way. I'm sure there is a more elegant solution for this. I'm thinking of having a global *Client object and accessing that through Unleash instead of initializing a NewClient every time. This way the user can initialize a global Client and add new ones as needed. In most cases one Client would be enough. At the least have a function to access the IsReady channel for the defaultClient instance.

ivarconr:
Sounds reasonable, but im not a go expert. What do think @jrbarron ?

shivid:
I made a Pull Request with a possible solution as well.

shividhar:
I made a Pull Request with a possible solution as well.

jrbarron:
@shivid I pushed a patch that should address your use case to this branch: https://github.com/Unleash/unleash-client-go/tree/feature/wait-for-ready Could you maybe check it out and give it a go before I merge?

jrbarron:
We now have a `WaitForReady` function in the `v3` branch which I think covers this use case so closing this issue.

PR #4, opened 2017-06-07
Added functionality so we know when globally accessible client is ready for polling.

shividhar:
Refs #3

### DIFF
--- a/unleash.go
+++ b/unleash.go
@@ -44,6 +44,12 @@ func Initialize(options ...ConfigOption) (err error) {
 	return
 }
 
+// IsReady is a channel operated method which specifies when the global unleash
+// client is ready
+func IsReady() bool {
+  return <-defaultClient.Ready()
+}
+
 // Close will close the default client.
 func Close() error {
 	return defaultClient.Close()


unleash.go
```diff
@@ -44,6 +44,12 @@ func Initialize(options ...ConfigOption) (err error) {
 	return
 }
 
+// IsReady is a channel operated method which specifies when the global unleash
+// client is ready
+func IsReady() bool {
```
jrbarron: @shividhar This is a perfectly valid use case, but I don't think this solution will work unfortunately. Since the function reads from a channel, it will block the calling routine until the channel is written to. I was thinking the quick fix would be that you simply rename the function to `WaitForReady()` which will clearly indicate that it will block until the Unleash client is ready. However, calling this function a second time (your use case) would then hang indefinitely because the channel is now empty and only written to once. Also if you register a `RepositoryListener`, then it will create a race condition with either the listener _or_ this `WaitForReady` receiving the value and the loser would then hang forever. Obviously this is suboptimal and can lead to non-intuitive use of the library.

This is a very good use case though and exposes a limitation in the Go implementation of the client that doesn't exist in the Javascript version because EventEmitters can have multiple listeners. I'm going to give this some more thought and see if I can come up with an elegant solution in the library. Feel free to do the same by updating this pull request if you have some ideas.

In the meantime, I think a good workaround for this would be making a struct that implements the `RepositoryListener` interface (`OnReady`) and sets a boolean to track that it has been initialized. You could then make your own version of `IsEnabled` that waits for the signal before calling the default client. Something like this:

```go
type UnleashWrapper struct {
	isReady bool
	onReady chan bool
}

func (h *UnleashWrapper) IsEnabled(feature string, options ...unleash.FeatureOption) bool {
	if !h.isReady {
		h.isReady = <-h.onReady // block until ready
	}
	return unleash.IsEnabled(feature, options)
}

func (h *UnleashWrapper) OnReady() {
	h.onReady <- true
}

var wrapper UnleashWrapper

func init() {
	wrapper = UnleashWrapper{onReady: make(chan bool, 1)}
	unleash.Initialize(
		unleash.WithListener(&wrapper),
		unleash.WithAppName("my-application"),
		unleash.WithUrl("http://unleash.herokuapp.com/api/"),
	)

	wrapper.IsEnabled("foo")
}
...
```

jrbarron:
Closing due to inactivity.

PR #6, opened 2017-06-26
Feature/custom headers

jrbarron:
This pull requests contains a few small fixes around headers. We were not currently setting the `User-Agent` header in the Go client so this fixes that and in addition adds support for custom headers. Also, there was a lot of duplicate code in the metrics class for POSTing to register the client and sending metrics so I factored that out into a separate function that marshals to JSON, sets the correct headers and sends the request.

### DIFF
--- a/client.go
+++ b/client.go
@@ -157,7 +157,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 	uc.metrics = newMetrics(
 		metricsOptions{
 			appName:         uc.options.appName,
-			instanceID:      uc.options.instanceId,
+			instanceId:      uc.options.instanceId,
 			strategies:      strategyNames,
 			metricsInterval: uc.options.metricsInterval,
 			url:             *parsedUrl,
--- a/client.go
+++ b/client.go
@@ -140,6 +140,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			refreshInterval: uc.options.refreshInterval,
 			storage:         uc.options.storage,
 			httpClient:      uc.options.httpClient,
+			customHeaders:   uc.options.customHeaders,
 		},
 		repositoryChannels{
 			errorChannels: errChannels,
@@ -162,6 +163,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			metricsInterval: uc.options.metricsInterval,
 			url:             *parsedUrl,
 			httpClient:      uc.options.httpClient,
+			customHeaders:   uc.options.customHeaders,
 		},
 		metricsChannels{
 			errorChannels: errChannels,
--- a/config.go
+++ b/config.go
@@ -21,6 +21,7 @@ type configOption struct {
 	listener        interface{}
 	storage         Storage
 	httpClient      *http.Client
+	customHeaders	http.Header
 }
 
 // ConfigOption represents a option for configuring the client.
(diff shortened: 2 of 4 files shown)


jrbarron:
@ivarconr I know you are not that familiar with Go, but this should be fairly straightforward and it's always good to have a second set of eyes on it :)

client.go
```diff
@@ -157,11 +158,12 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 	uc.metrics = newMetrics(
 		metricsOptions{
 			appName:         uc.options.appName,
-			instanceID:      uc.options.instanceId,
+			instanceId:      uc.options.instanceId,
```
ivarconr: oh my, yes!

metrics.go
```diff
@@ -209,6 +182,29 @@ func (m *metrics) sendMetrics() {
...
+		return nil, err
+	}
+
+	req, err := http.NewRequest("POST", url.String(), &body)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Add("Unleash-Appname", m.options.appName)
+	req.Header.Add("Unleash-Instanceid", m.options.instanceId)
```
ivarconr: java/node use uppercase for these headers ("UNLEASH-APPNAME" and "UNLEASH-INSTANCEID"). 

Might be a bad choice and not following the naming convention for headers?

ivarconr:
@jrbarron I would love to learn, maybe you can recommend a crash course to get me started?

config.go
```diff
@@ -21,6 +21,7 @@ type configOption struct {
 	listener        interface{}
 	storage         Storage
 	httpClient      *http.Client
+	customHeaders	http.Header
```
ivarconr: I would expect that you had to specify this as an array? 

`[] http.Header` ?

ivarconr:
Would it be possible to write a test for it also, verifying that the headers are sent as part of the request? In Java I tend to use WireMock for such tests. Is there anything similar for Go that works? Maybe [httpmock](https://github.com/jarcoal/httpmock)?

config.go
jrbarron: A very reasonable expectation, but `http.Header` is a [builtin](https://golang.org/pkg/net/http/#Header) type which is defined as `map[string][]string`. That's a bit ugly but it means that the `Header` type is a map of string arrays. In Javascript, it would look like this:

```js
const header = {
  'Foo': ['bar', 'baz']
}
```
I don't really know why they chose to use an array for the values instead of just a string, but the HTTP spec says multiple headers can have the same name if the values can be combined so maybe that is why. Rarely used though.

metrics.go
jrbarron: I'll change this to be consistent with the other clients. Go will internally send the canonical form with the request, but I guess the server handles this.

jrbarron:
I investigated this a bit while making this change and I agree, we should use some mocking library here and get some better coverage. I've never used one before so I'll do some research and see if I can get a test added before merging.

jrbarron:
@ivarconr Dave Cheney (Ubuntu developer) has a very thorough list of resources for new Go developers: https://dave.cheney.net/resources-for-new-go-programmers

I would personally recommend starting here: https://tour.golang.org/welcome/1

If you get through that and are hungry for more, there is [Effective Go](https://golang.org/doc/effective_go.html). As always though, the best way to learn this stuff is just to jump in so next time we need something implemented in the Go client, I nominate you to have the first crack at it! 😉

PR #8, opened 2017-07-19
taking the parameters from the rule instead of the empty strategy

itairegev:
the parameters are taken from the wrong place, and are always empty


SimenB:
Thanks!

Could you add a test for this?

/cc @jrbarron

jrbarron:
I'll merge this without a test for now since it is quite a serious bug. I have started improving test coverage in a separate branch called `improvements/tests` so if a test were to be added, it might be better to use that branch since it adds some dependencies which would help with this.

PR #9, opened 2017-07-30
metrics channel never read

itairegev:
the metrics channel was never read, only written to, this caused the app to stop responding.

### DIFF
--- a/client.go
+++ b/client.go
@@ -239,7 +239,8 @@ func (uc Client) IsEnabled(feature string, options ...FeatureOption) (enabled bo
 			// TODO: warnOnce missingStrategy
 			continue
 		}
-		if foundStrategy.IsEnabled(s.Parameters, opts.ctx) {
+
+    if foundStrategy.IsEnabled(s.Parameters, opts.ctx) {
 			return true
 		}
 	}
--- a/client.go
+++ b/client.go
@@ -140,6 +140,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			refreshInterval: uc.options.refreshInterval,
 			storage:         uc.options.storage,
 			httpClient:      uc.options.httpClient,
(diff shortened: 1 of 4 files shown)


jrbarron:
@itairegev Hmm, I don't think removing the line where we write to the channel is the correct fix. This will mean that users can never be notified when the client has successfully sent the metrics which is not we want since the other versions of the client support this. You mention that the channel is never read from, but that is not true because the client's `sync()` function will read from the channel when it reads from all the other channels. Note that the `sync` function will only be called if you create the client with a Listener as described here:

https://godoc.org/github.com/Unleash/unleash-client-go#hdr-Using_the_Listener_Interfaces

If you do not supply a listener then it is expected that you yourself read from ALL the channels to prevent the app from locking up. In retrospect, I may have implemented this differently such that a listener would not be required, but my fear was that this would make it rather easy for people to ignore errors coming from the client and this practice is frowned upon in the Go community. You can easily get this behaviour yourself though by just doing something like calling `WithListener(false)`, but I strongly recommend passing in a listener that implements the `ErrorListener` interface so you at least know when errors have occured.

itairegev:
having listeners is recommended, but if it is possible to run with no listeners, it can't get locked, i agree commenting this line is not the right solution, but it is a fix till a solution is found.

jrbarron:
@itairegev The only way you can run without listeners is that if you take the responsibility of draining the channels (all of them) yourself. This option was provided so that the Unleash client could integrate in apps that already ran an application loop, but if you do not do this, then you really should be using the listeners. Unless I am misunderstanding something, you should be able to get the same behavior as your PR, just by calling `WithListener(false)`, but then you are essentially ignoring errors at your own peril :) Does my proposed workaround not work for you?

jrbarron:
Closing due to inactivity.

Issue #18, opened 2018-04-26
Add makefile and dev-section in readme


nilsmagnus:
Add makefile and a dev-section in readme on how to contribute

PR #19, opened 2018-04-26
Add makefile and update README. This solves #18

nilsmagnus:


README.md
```diff
@@ -74,3 +74,25 @@ ctx := context.Context{
...
+
+Run tests:
+
+    make 
+    
+Run lint check:
+
+    make lint
+    
+Run code-style checks:(currently failing)
```
jrbarron: It would be nice if all the checks were passing before merging this, but I haven't had time lately to fix this so I can just merge it for now and we can create issues for fixing these things.

Makefile
```diff
@@ -0,0 +1,49 @@
...
+# The name of the executable (default is current directory name)
+TARGET := $(shell echo $${PWD\#\#*/})
+.DEFAULT_GOAL: $(TARGET)
+
+# These will be provided to the target
+VERSION := 1.0.0
+BUILD := `git rev-parse HEAD`
+
+# Use linker flags to provide version/build settings to the target
+LDFLAGS=-ldflags "-X=main.Version=$(VERSION) -X=main.Build=$(BUILD)"
```
jrbarron: I guess this a fairly standard Makefile for Go, but is there any point in including these when we don't have a `cmd` binary? Not a showstopper, just asking.. 😉

nilsmagnus: You are right, I copied it from a different project, so it can be omitted.

README.md
nilsmagnus: I think a separate issue for the actual fixes would be cleaner :)

nilsmagnus:
I see that the tests are failing as well, this project has probably not been touched for a while?

jrbarron:
The tests use the public Unleash demo server (obviously suboptimal) which of course can be edited so some of the feature toggles that we depend on might have disappeared. Would be better to mock this, I just haven't gotten around to it yet 😄

Makefile
jrbarron: @nilsmagnus Did you want to remove these lines? Then I think we can merge this.

nilsmagnus:
I finally got around to update the makefile now.

jrbarron:
@nilsmagnus Cool! I just pulled your branch to give it a try, but I got an error because the commands under `get-deps` are prefixed with spaces instead of tabs so make barfs. That should be a quick fix :)

One more thing I noticed, the Makefile doesn't install the test dependencies so the default target fails. I'm wondering if it would be good to have the following line in the `build` target:

`go get -t ./...`

What do you think?

Finally I don't think we need to delete *.png files in the clean target since we don't have any so you can probably remove that line too ;)

nilsmagnus:
Agree on your comments. Regarding dependencies, 'vgo' is still in the making, but when it is finally released you should probably consider to move to it. I have been using 'dep' for a while and am satisfied, but I would prefer 'vgo' when it is released as a part of the go-toolchain.

PR #23, opened 2018-06-12
Osio sync issue

xcoulon:
Move the timer initialization in the loop, so a new timer is used after each sync, and a new event is triggered. Otherwise, the `sync` rountine halts after the first sync.


xcoulon:
sorry, wrong PR

PR #29, opened 2018-09-14
Testing/implement client spec test

jrbarron:
Looks like the Travis build is failing despite everything passing locally so I'll try to look at that when I get a chance.

Besides that, all of the client spec tests are now passing. I only had to fix one bug 1e6f2ea which was a pleasant surprise.

The way the test runner works is as follows:
 - Create and run the suite in `TestClientSpecificationSuite`
 - This implicitly calls `SetupTest()`
 - Which loads all the files and parses the JSON into an array of `TestDefinitions`
 - Each TestDefinition is executed which involves:
   - Create a mocked client
   - Iterate through the test cases and run them, passing them the mocked client
   - Compare results

### DIFF
--- a/client.go
+++ b/client.go
@@ -241,6 +241,10 @@ func (uc Client) IsEnabled(feature string, options ...FeatureOption) (enabled bo
 		return false
 	}
 
+	if len(f.Strategies) == 0 {
+		return f.Enabled
+	}
+
 	for _, s := range f.Strategies {
 		foundStrategy := uc.getStrategy(s.Name)
 		if foundStrategy == nil {
--- a/client.go
+++ b/client.go
@@ -36,10 +36,12 @@ type Client struct {
 	metricsListener    MetricListener
 	repositoryListener RepositoryListener
 	ready              chan bool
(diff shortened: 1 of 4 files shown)


ivarconr:
looks like https://travis-ci.org/Unleash/unleash-client-go/builds/428832620#L495 is missing a mock for `/client/features'

ivarconr:
Cool that you managed to use the client-specs :100:  . Will be an awesome tool when implementing support for 4.x (grouped strategies). :+1:

client.go
```diff
@@ -299,3 +307,11 @@ func (uc Client) getStrategy(name string) strategy.Strategy {
 	}
 	return nil
 }
+
+// WaitForReady will block until the internal repository has loaded the feature toggles from the
+// storage engine. It will return immediately if the repository is already ready.
+func (uc *Client) WaitForReady() {
+	if !uc.isReady {
```
ivarconr: great improvement!

jrbarron:
Agreed, I'll fix Travis before merging this one.

PR #31, opened 2018-11-20
metrics: Close http response bodies to avoid leaks

jallen-asapp:
In order to avoid a goroutine leak (the read and write loops on a persistent http connection), we need to close the response body every time.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -144,6 +144,7 @@ func (m *metrics) registerInstance() {
 		m.err(err)
 		return
 	}
+	resp.Body.Close()
 
 	if resp.StatusCode < http.StatusOK || resp.StatusCode > http.StatusMultipleChoices {
 		m.warn(fmt.Errorf("%s return %d", u.String(), resp.StatusCode))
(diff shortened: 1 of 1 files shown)


metrics.go
```diff
@@ -144,6 +144,7 @@ func (m *metrics) registerInstance() {
 		m.err(err)
 		return
 	}
+	resp.Body.Close()
```
jrbarron: Ah good catch. I forgot about this since we don't actually read the response body. I think we should maybe put a `defer` in front of this call though? Otherwise if the previous line enters the error case and returns, we still have the same problem.

metrics.go
```diff
@@ -172,6 +173,7 @@ func (m *metrics) sendMetrics() {
...
```
jrbarron: Same thing here. I think it makes sense to use `defer`.

metrics.go
jallen-asapp: Sure. There is probably no harm in deferring this. It could prevent unintended problems if we were ever to read from the body.

Issue #17, opened 2018-01-25
Another deadlock in metrics


yukw777:
Hello everyone, I've found another deadlock in the metrics event loop. It happens if the unleash server (`./client/metrics`) returns 404. Erroneous config value was the culprit for me.

1. The client attempts to send metrics but gets 404, so it tries to stop.
https://github.com/Unleash/unleash-client-go/blob/be883ea1af714aff427d0668dbd4fc549d4f701c/metrics.go#L176-L180

2. `stop()` sends `true` to `m.stopped` channel, and blocks.
https://github.com/Unleash/unleash-client-go/blob/be883ea1af714aff427d0668dbd4fc549d4f701c/metrics.go#L106

3. At this point, `sync()` is also blocked, since `sendMetrics()` is blocked. But `sync()` is the one that could free everything up by taking the value from `m.stopped`. Hence, deadlock.
https://github.com/Unleash/unleash-client-go/blob/be883ea1af714aff427d0668dbd4fc549d4f701c/metrics.go#L124-L129

jrbarron:
After merging #52 this issue seems to be resolved.

PR #32, opened 2019-01-18
Prevent metrics send from blocking count and isEnabled

jallen-asapp:
In short, all `IsEnabled` calls currently wait for active send metrics calls to complete before returning.
With any sort of temporary outage or unavailability of the unleash service, clients will experience potentially significant delays in all feature flag evaluations.

metrics: prevent send from blocking count

Changes from using the channel mechanism to protect the buckets to using
a mutex. This reduces the worst-case cost of counting on each flag check
with a slight increase in the best-case cost.

Old best case:
* sync loop is waiting to receive. count without blocking.
New best case:
* no lock contention, key exists in bucket. fetch, branch, set in map.

Old worst case:
* send is deadlocked after a 404. count blocks forever, blocking
IsEnabled calls.
* send is taking a long time (DNS issues, etc.). count blocks until http
timeout.
New worst case:
* large number of routines contending for lock, including send. count
blocks until the simple in-process operations complete.

related to #17, in that this prevents a send deadlock from deadlocking a full application.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/Unleash/unleash-client-go/v3/internal/api"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
+
+	"github.com/Unleash/unleash-client-go/v3/internal/api"
 )
 
 // MetricsData represents the data sent to the unleash server.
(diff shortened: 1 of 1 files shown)


jrbarron:
@jallen-asapp Thanks for your contribution! At a first glance this looks reasonable, but I want to do a bit of manual testing first to double check there are not side effects. I'll try to get back to as soon as possible so we can get this merged.

jallen-asapp:
We ran some load tests with this change in our load testing environment.

We confirmed that with unleash unavailable, and no fix, we experience an outage. We also confirmed that with unleash unavailable, and the fix, we don't experience an outage.

We're not in a tremendous rush, but I wanted to share that we've put this through load testing.

PR #33, opened 2019-01-24
Fix 'WithDisableMetrics' to actually stop sending metrics

### LINKED ISSUE #15: Deadlock in metrics
There's a deadlock in how the go client handles metrics data, specifically when it handles the "sent" event. There are other issues in the codebase that were uncovered during debugging, so I'll list all of them here for now, but I'm happy to make a separate issue for each if necessary.

The deadlock happens when you don't pass in an "optional" event listener (`WithListener`) when initializing the client. The unbuffered sent channel receives a value
https://github.com/Unleash/unleash-client-go/blob/c586a9e5f31e9788ba0dc3f0726150717330d971/metrics.go#L186
, but b/c the event listener is nil, the receiver is never started
https://github.com/Unleash/unleash-client-go/blob/c586a9e5f31e9788ba0dc3f0726150717330d971/client.go#L97-L110
, so the next time a payload is sent to the sent channel, it blocks.

There are some other issues related to this bug.

1. `WithDisableMetrics` doesn't work, b/c the value doesn't get passed into metrics from the clinet.
https://github.com/Unleash/unleash-client-go/blob/c586a9e5f31e9788ba0dc3f0726150717330d971/client.go#L163-L171
1. If you pass in the value down to metrics, you get a nil pointer exception b/c the timer is never initialized.
https://github.com/Unleash/unleash-client-go/blob/c586a9e5f31e9788ba0dc3f0726150717330d971/metrics.go#L95-L99
1. Setting metrics interval to 0 doesn't work either since it now blocks on other channels.
https://github.com/Unleash/unleash-client-go/blob/c586a9e5f31e9788ba0dc3f0726150717330d971/metrics.go#L80-L84
1. Metrics is too tightly coupled to the client. Proper dependency inversion would help here.

The only true workaround that works right now is to set a listener, even if it's a noop.

Happy to help out in anyway I can. Thanks!

jrbarron:
This was previously broken because we were never passing down the
boolean value from the client to the metrics struct. Doing so implies
that the timer is never created so we can't read from the channel in the
sync function. If metrics are disabled, there is really no reason to
even run the metrics routine, so just return instead.

Fixes #15

### DIFF
--- a/client.go
+++ b/client.go
@@ -171,6 +171,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			url:             *parsedUrl,
 			httpClient:      uc.options.httpClient,
 			customHeaders:   uc.options.customHeaders,
+			disableMetrics:  uc.options.disableMetrics,
 		},
 		metricsChannels{
 			errorChannels: errChannels,
--- a/client.go
+++ b/client.go
@@ -91,6 +91,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 		count:         make(chan metric),
 		sent:          make(chan MetricsData),
 		registered:    make(chan ClientData, 1),
+		closed:        make(chan bool),
 	}
 
 	for _, opt := range options {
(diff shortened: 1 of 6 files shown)


jrbarron:
Fixed by PR #51

PR #37, opened 2019-04-26
Fix a few godoc typos

tormoder:


tormoder:
Go 1.8 failures are due to `github.com/spaolacci/murmur3` now are using `math/bits` introduced in Go 1.9.

PR #38, opened 2019-04-26
Bump Travis Go version to 1.12

tormoder:


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - "1.8"
-  - "1.11"
+  - 1.8.x
```
jrbarron: Since you mentioned the build fails on 1.8 due to murmur dependencies, do you think we should bump this to 1.9?

tormoder: Yes, I think so.

PR #39, opened 2019-04-26
Simplify WaitForReady implementation

tormoder:
The channel type could also be changed to `chan struct{}{}` to show that it is only used for signaling.


frekui:
Is there a reason for why this hasn't been merged yet? It simplifies the code and also fixes a race condition in `WaitForReady` if two goroutines call it concurrently.

PR #40, opened 2019-04-26
[v4] Remove unneeded parameter from WithDisableMetrics

tormoder:


tormoder:
Sorry, I forgot that this is a breaking change, and probably can't be merged for the `v3` branch.

jrbarron:
I have started a v4 branch with some other breaking changes, but haven't pushed it yet. I'll see if I can push that now and then you can just change the target branch.

jrbarron:
Ok, I pushed to `wip/v4` since there are some issues with the v4 branch at the moment so I didn't want to give the impression that it is finished ;) I have been marking commits that break the API with a `[API CHANGE]` so it will be easier to find those commits when we need to write a a migration guide. Would be great if you could amend your commit to do the same.

The change itself makes sense I think.

tormoder:
Thanks :-)

I'll close this PR, and use the `wip/v4` branch as base instead (sorry for the noise).

PR #43, opened 2019-06-25
disableMetrics does not actually disable metrics reporting in v3

davidsiefert:
Was getting a deadlock in metrics reporting, so I tried to disable metrics but it did not work.  Here is a patch to fix it:

Needed to propagate disableMetrics option from client to the metrics options block to disable metrics from the beginning of execution. Once disabled, `sync()` panics because the timer is not initialized in the case that metrics are disabled, so I added a check to return early when disableMetrics is set.

### DIFF
--- a/client.go
+++ b/client.go
@@ -171,6 +171,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			url:             *parsedUrl,
 			httpClient:      uc.options.httpClient,
 			customHeaders:   uc.options.customHeaders,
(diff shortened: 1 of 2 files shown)


ivarconr:
I believe this is a duplicate of #33. What do you think @jrbarron ?

jrbarron:
Yes, this looks like a subset of PR #33, but I can merge it and then rebase my PR on top.

ivarconr:
great! :dango:

PR #47, opened 2019-08-16
storage: use filepath.Join when constructing storage path

### LINKED ISSUE #46: Default backup path doesn't work on Linux
By default `NewClient` uses `getTmpDirPath` as the backup path. `getTmpDirPath` returns whatever `os.TempDir` returns. On Linux (tested on Ubuntu 18.04) `os.TempDir` returns a path without a trailing slash. In `storage.go` the final path is constructed using `fmt.Sprintf("%sunleash-repo-schema-v1-%s.json", backupPath, appName)` as there is no slash between the backup path and the rest of the path the file creation will typically fail with a permission denied error.

tormoder:
Fixes #46.

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 )
 
 // Storage is an interface that can be implemented in order to have control over how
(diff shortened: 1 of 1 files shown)


frekui:
I have verified that this change fixes the problem for me. Good job!

PR #51, opened 2019-09-14
Make go test -race ./... pass

frekui:
This PR makes Client.Close return *after* all resources have been freed
that's used by the client. In particular, when Client.Close returns all
goroutines started by the client have exited, any errors and warnings reported
by the client have been sent to their respective channels or listeners, and no
HTTP requests started by the client will still be in flight.

https://github.com/Unleash/unleash-client-go/issues/17 isn't fixed by this change, but I believe it's a step in the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,6 @@ Run code-style checks:(currently failing)
     
     make strict-check
     
-Run race-tests(currently failing):
+Run race-tests:
  
     make test-all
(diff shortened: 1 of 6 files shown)


frekui:
Turns out this needs some more work. I will reopen later.

jrbarron:
@frekui Great work and thanks for the PR! I see that this includes the fixes for disable metrics as well. I never got around to merging that one because the tests failed sporadically in Travis, but they seem to be working well on your branch so I think this is safe to merge. I did some testing locally and everything seems to work.

PR #52, opened 2019-09-28
Report all metrics even if /client/metrics fails temporarily

frekui:
Before this change metrics were reported as sent even if `/client/metrics` returned a non-successful status code. Furthermore, the metrics were reset in this case thus some metrics were lost. This PR fixes the issue by only reporting successful POSTs to `/client/metrics` as sent and also metrics from failed POSTs are sent in the next POST to `/client/metrics`.

I decided to use httptest instead of gock for the test. It seemed simpler in this case.

As a side-effect this PR fixes https://github.com/Unleash/unleash-client-go/issues/17 as the special casing of 404 no longer exist.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -61,7 +61,7 @@ type metrics struct {
 	started  time.Time
 	bucketMu sync.Mutex
 	bucket   api.Bucket
-	timer    *time.Timer
+	ticker   *time.Ticker
 	close    chan struct{}
 	closed   chan struct{}
 	ctx      context.Context
@@ -84,15 +84,12 @@ func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
 		m.options.httpClient = http.DefaultClient
 	}
 
-	m.bucketMu.Lock()
 	m.resetBucket()
-	m.bucketMu.Unlock()
-
 	if m.options.metricsInterval <= 0 {
 		m.options.disableMetrics = true
 	}
 	if !m.options.disableMetrics {
-		m.startTimer()
+		m.ticker = time.NewTicker(m.options.metricsInterval)
 		m.registerInstance()
 		go m.sync()
 	}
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -146,36 +139,43 @@ func (m *metrics) registerInstance() {
...
-	if resp.StatusCode == http.StatusNotFound {
-		m.warn(fmt.Errorf("%s return 404, stopping metrics", u.String()))
-		m.Close()
-		return
-	}
-
 	if resp.StatusCode < http.StatusOK || resp.StatusCode > http.StatusMultipleChoices {
 		m.warn(fmt.Errorf("%s return %d", u.String(), resp.StatusCode))
-	}
+		// The post failed, re-add the metrics we attempted to send so
```
ivarconr: This is quite clever, I think it can be reused in other client SDKs as-well!

metrics.go
```diff
@@ -146,36 +139,43 @@ func (m *metrics) registerInstance() {
...
-	m.startTimer()
 	resp, err := m.doPost(u, payload)
-
 	if err != nil {
 		m.err(err)
 		return
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == http.StatusNotFound {
```
ivarconr: What happens if the server deliberately does not handle metrics? Will we then accumulate metrics for-ever? Could that end up being a memory leak? 

From a user-perspective they should probably use the client with `disableMetrics=true` option set, maybe we should inform the user in the error message?

jrbarron: No memory leak since we only ever store two `int32` in the bucket, but I guess theoretically it could overflow to be negative numbers at some point. Hopefully the server can handle those, but this is an edge case I think.

We will broadcast a warning every time the metrics fail to upload so the user does get informed to some extent.

frekui:
Thanks for reviewing and merging. Good point about splitting into smaller commits.

PR #56, opened 2019-10-10
Implement support for FlexibleRolloutStrategy and Constraints

jrbarron:
This PR implements the new FlexibleRolloutStrategy and also adds support for Constraints. The PR still needs a few more things before it is considered finished:

- [x] Waiting on #53 to be merged so this branch can be rebased on top
- [ ] Unit tests for FlexibleRolloutStrategy
- [x] Waiting for https://github.com/Unleash/client-specification/pull/6 to be merged so the client specification tests will pass

### DIFF
--- a/internal/api/constraint.go
+++ b/internal/api/constraint.go
@@ -1,0 +1,14 @@
+package api
+
+type Operator string
+
+const (
+	OperatorIn    Operator = "IN"
+	OperatorNotIn Operator = "NOT IN"
+)
+
+type Constraint struct {
+	ContextName string   `json:"contextName"`
+	Operator    Operator `json:"operator"`
+	Values      []string `json:"values"`
+}
--- a/internal/api/constraint.go
+++ b/internal/api/constraint.go
@@ -4,7 +4,7 @@ type Operator string
 
 const (
 	OperatorIn    Operator = "IN"
-	OperatorNotIn Operator = "NOT IN"
+	OperatorNotIn Operator = "NOT_IN"
 )
 
 type Constraint struct {
--- a/internal/api/constraint.go
+++ b/internal/api/constraint.go
@@ -1,12 +1,20 @@
 package api
 
+// Operator is a type representing a constraint operator
 type Operator string
 
 const (
-	OperatorIn    Operator = "IN"
+	// OperatorIn indicates that the context values must be
+	// contained within those specified in the constraint.
+	OperatorIn Operator = "IN"
+
+	// OperatorNotIn indicates that the context values must
+	// NOT be contained within those specified in the constraint.
 	OperatorNotIn Operator = "NOT_IN"
 )
 
+// Constraint represents a constraint on a particular context
+// value.
 type Constraint struct {
 	ContextName string   `json:"contextName"`
 	Operator    Operator `json:"operator"`
--- a/internal/strategies/flexible_rollout.go
+++ b/internal/strategies/flexible_rollout.go
@@ -1,9 +1,6 @@
 package strategies
 
 import (
-	"math/rand"
-	"strconv"
-
 	"github.com/Unleash/unleash-client-go/v3/context"
 	"github.com/Unleash/unleash-client-go/v3/strategy"
 )
@@ -18,23 +15,19 @@ const (
 )
 
 type flexibleRolloutStrategy struct {
-	random *rand.Rand
+	random *rng
 }
 
 // NewFlexibleRolloutStrategy creates a new instance of the flexible rollout strategy.
 func NewFlexibleRolloutStrategy() *flexibleRolloutStrategy {
 	s := &flexibleRolloutStrategy{
-		random: newRand(),
+		random: newRng(),
 	}
 	return s
 }
 
 func (s flexibleRolloutStrategy) Name() string {
 	return "flexibleRollout"
-}
-
-func (s flexibleRolloutStrategy) randomNumber() string {
-	return strconv.Itoa(s.random.Intn(100) + 1)
 }
 
 func (s flexibleRolloutStrategy) resolveStickiness(st stickiness, ctx context.Context) string {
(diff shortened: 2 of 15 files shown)


internal/strategies/flexible_rollout.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+type flexibleRolloutStrategy struct {
+	random *rand.Rand
+}
+
+// NewFlexibleRolloutStrategy creates a new instance of the flexible rollout strategy.
+func NewFlexibleRolloutStrategy() *flexibleRolloutStrategy {
+	s := &flexibleRolloutStrategy{
+		random: rand.New(rand.NewSource(time.Now().Unix())),
```
frekui: It would be better to use e.g. `time.Now().UnixNano() + int64(os.Getpid())` here (this is what ioutil.TempFile is using to seed it's RNG). Alternatively, crypto/rand can be used to get a proper random seed. With the code in the PR there is a risk that several instances of an application get the same seed if they are all restarted at approximately the same time (e.g., during a deploy of a new version).

internal/api/constraint.go
```diff
@@ -0,0 +1,14 @@
+package api
+
+type Operator string
+
+const (
+	OperatorIn    Operator = "IN"
+	OperatorNotIn Operator = "NOT IN"
```
ivarconr: Is this a typo? In the protocol you will get `NOT_IN`, see https://github.com/Unleash/client-specification/pull/5/files#diff-07b98f0d3424fd4c9547b512848b9521R67

jrbarron: Wasn't really a typo since I read it in the Google Docs for the Constraints protocol you sent:
https://docs.google.com/document/d/1WxdODm1xKLaArucD7E5nHg4nruftRfkaWibaGV32GaI/edit
I'll fix mine if you fix yours 😂

ivarconr: Oh this is embarrassing, fixed now.

internal/strategies/flexible_rollout.go
ivarconr: Why not just use math/rand? Already used in [gradual_rollout_random.go](https://github.com/Unleash/unleash-client-go/blob/master/internal/strategies/gradual_rollout_random.go#L36)

jrbarron: Fair point, but same seed is not really an issue is it? The number is not used for anything cryptographic and `crypto/rand` is an order of magnitude slower. Right now, the implementation here is similar to Java:

```java
random = new Random();
```

And Node

```js
Math.floor(Math.random() * 100) + 1
```

Happy to change this if there is consensus, but I feel performance is more important than unique seed here?

internal/strategies/flexible_rollout.go
```diff
@@ -0,0 +1,79 @@
...
+func (s flexibleRolloutStrategy) resolveStickiness(st stickiness, ctx context.Context) string {
+	switch st {
+	case userIDStickiness:
+		return ctx.UserId
+	case sessionIDStickiness:
+		return ctx.SessionId
+	case randomStickiness:
+		return s.randomNumber()
+	default:
+		return coalesce(ctx.UserId, ctx.SessionId, s.randomNumber())
```
ivarconr: Nice, I did not now that `coalesce` was a thing!

internal/strategies/flexible_rollout.go
ivarconr: +1 for performance.

internal/strategies/flexible_rollout.go
jrbarron: It's not builtin unfortunately so I borrowed the concept from Postgres and implemented it in the "helpers" file. I think I might want to make a public API for these helpers since it makes writing custom strategies API. Especially things like `normalizedValue` are very helpful so this has already been requested by one user.

ivarconr: Yes indeed. I'm planning the same thing for other clients.

internal/strategies/flexible_rollout.go
ivarconr: Btw, allowing to the seed will enable unit testing of the implementation.

frekui: No, the code isn't similar to the Java or Node code. The Java doc says
> This constructor sets the seed of the random number generator to a value very likely to be distinct from any other invocation of this constructor.

From what I understand of the JS documentation `Math.random()` returns a uniformly distributed number if the range [0, 1).

Unless I have missed something `NewFlexibleRolloutStrategy` is only called once, if so using crypto/rand won't be a problem. But, sure, if you're worried about performance use `time.Now().UnixNano() + int64(os.Getpid())` instead.

internal/strategies/helpers.go
```diff
@@ -64,3 +67,8 @@ func coalesce(str ...string) string {
 	}
 	return ""
 }
+
+// newRand creates a new random number generator
+func newRand() *rand.Rand {
+	return rand.New(rand.NewSource(time.Now().Unix() + int64(os.Getpid())))
```
frekui: This is an improvement, but can you please change `Unix()` to `UnixNano()`?

internal/strategies/flexible_rollout.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return s
+}
+
+func (s flexibleRolloutStrategy) Name() string {
+	return "flexibleRollout"
+}
+
+func (s flexibleRolloutStrategy) randomNumber() string {
+	return strconv.Itoa(s.random.Intn(100) + 1)
```
frekui: There is a data race here. It's not safe to call methods on rand.Rand from more than one goroutine concurrently. If a user of the module calls Client.IsEnabled from multiple goroutines concurrently then we might end up in this function in more than one goroutine concurrently. There is a similar data race in gradualRolloutRandomStrategy.

jrbarron: Good catch, pushed a fix now.

PR #62, opened 2019-10-24
Add a fallback function option to IsEnabled

### LINKED ISSUE #60: feat: add support for fallback function
Today it is possible to define a fallback value which the `IsEnabled` will return if the feature toggle is not defined. 

We want to also allow the user to send in a fallbackFunction instead of a fallback value (which will take precedence over fallback value).

The fallback function should get to arguments injected: `toggleName` and `unleashContext`. 

Example on how it would work:

```go
unleash.IsEnabled("foobar", unleash.WithFallbackFunction(myFunc))
```

If the user injects both a fallback value and fallbackFunction, like the example below, the result should be to only call the fallback function in the scenario that the toggle is not defined:

```go
unleash.IsEnabled("foobar", unleash.WithFallback(true), unleash.WithFallbackFunction(myFunc))
```

jrbarron:
This allows users to specify a dynamic fallback instead of falling back
to a fixed boolean value.

Fixes #60

TODO:

- [x] Unit tests

### DIFF
--- a/config.go
+++ b/config.go
@@ -129,7 +129,7 @@ func WithCustomHeaders(headers http.Header) ConfigOption {
 	}
 }
 
-// FallbackFunc represents a function to be called if the feature if not found.
+// FallbackFunc represents a function to be called if the feature is not found.
 type FallbackFunc func(feature string, ctx *context.Context) bool
 
 type featureOption struct {
--- a/config.go
+++ b/config.go
@@ -129,9 +129,13 @@ func WithCustomHeaders(headers http.Header) ConfigOption {
 	}
 }
 
+// FallbackFunc represents a function to be called if the feature if not found.
+type FallbackFunc func(feature string, ctx *context.Context) bool
+
 type featureOption struct {
-	fallback *bool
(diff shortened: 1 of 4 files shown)


ivarconr:
Nice, looks good! 

Maybe also a note in the README.md on how to use it?

config.go
```diff
@@ -129,9 +129,13 @@ func WithCustomHeaders(headers http.Header) ConfigOption {
 	}
 }
 
+// FallbackFunc represents a function to be called if the feature if not found.
```
frekui: Nit: "if not found" -> "is not found".

ivarconr:
Whats left before we can merge this one?

jrbarron:
I need to fix the typo Fredrik found, update the README and ideally write a unit test, but I guess that is not a blocker for merging. I'll sit down now and finish it.

config.go
jrbarron: Fixed.

jrbarron:
I added an example now, but didn't update the README since it didn't really have any documentation apart from installation instructions.

PR #63, opened 2019-10-25
TestMetrics_SendMetricsFail: Fix non-deterministic test failure

frekui:
The test failure can be made deterministic by adding this sleep:

    diff --git a/metrics_test.go b/metrics_test.go
    index 68345c3..47ea377 100644
    --- a/metrics_test.go
    +++ b/metrics_test.go
    @@ -206,6 +206,7 @@ func TestMetrics_SendMetricsFail(t *testing.T) {
            // /client/metrics returns 400, check that the counts aren't reset.
            m.count("foo", true)
            ck(400, 1, 0, <-metricsCalls)
    +       time.Sleep(100 * time.Millisecond)
            m.count("foo", false)
            ck(400, 1, 1, <-metricsCalls)
            m.count("foo", true)

The body sent to /client/metrics contains two timestamps. The de-duplication
done in the HTTP handler in the test therefore didn't work as expected. As a
fix, parse the request body in the HTTP handler and compare only the toggles
when doing the de-duplication.

### DIFF
--- a/metrics_test.go
+++ b/metrics_test.go
@@ -1,12 +1,12 @@
 package unleash
 
 import (
-	"bytes"
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"reflect"
 	"strings"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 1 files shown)


frekui:
The test failure on Go 1.9 seems to be due to #61.

frekui:
Bumping to Go 1.10 in Travis sounds good to me. Go 1.12 should probably be bumped to 1.13 so that the client library is tested on the latest stable version of Go.

tormoder:
Would it make sense to specify and state the supported versions somewhere? (e.g. in the README).
And then also only use those versions for CI.

For example, only the two latest releases are maintained by the Go team.

jrbarron:
Yes, it probably make sense to at least mention which versions we test against. I can update the README.

PR #66, opened 2019-10-28
Cleanup Makefile

frekui:
I would be happy to remove the `$(TARGET)`, `build`, and `clean` targets as well. Are they used for anything?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,48 +1,28 @@
 SHELL := /bin/bash
 
-# The name of the executable (default is current directory name)
-TARGET := $(shell echo $${PWD\#\#*/})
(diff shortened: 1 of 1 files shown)


jrbarron:
The Makefile has largely been added and maintained by other members of the community so I can't really present an argument for why things are the way they are, but your changes make sense to me :)

The whole target/build/clean stuff seems a bit unnecessary since this is just a library so I can see no reason to keep it. Feel free to remove it and we can merge this.

frekui:
Updated. The `$(TARGET)`, `build`, and `clean` have been removed now.

PR #70, opened 2020-03-13
Added public API for listing all feature toggles

codebutler:


jrbarron:
@codebutler Thanks for the contribution! I'll review the implementation shortly ;)

@ivarconr What do you think about listing the features as a general concept? Do the other clients support this already?

repository.go
```diff
@@ -150,6 +150,14 @@ func (r *repository) getToggle(key string) *api.Feature {
 	return nil
 }
 
+func (r *repository) list() []api.Feature {
+	var features []api.Feature
```
jrbarron: You should probably lock the mutex for reading here before reading from the store. This should just be a matter of adding:

```go
r.RLock()
defer r.RUnlock()
```
to the top of the function.

feature/feature.go
```diff
@@ -0,0 +1,7 @@
+package feature
+
+type Feature struct {
```
jrbarron: The struct itself and the fields should be documented since they are exported.

ivarconr:
> @ivarconr What do you think about listing the features as a general concept? Do the other clients support this already?

Yes, we have seen that it from time to time are valid use cases where one would like to get the full feature toggle definitions. In the node client you can use the `getFeatureToggleDefinitions()`:

https://github.com/Unleash/unleash-client-node#3-use-unleash

client.go
```diff
@@ -347,3 +348,16 @@ func (uc *Client) getStrategy(name string) strategy.Strategy {
...
+
+// ListFeatures returns all available features toggles.
+func (uc *Client) ListFeatures() []*feature.Feature {
+	var features []*feature.Feature
+	for _, f := range uc.repository.list() {
+		features = append(features, &feature.Feature{
+			Name:         f.Name,
+			Description:  f.Description,
+			Enabled:      f.Enabled,
+		})
```
ivarconr: should also include all details (strategy and variant details missing)

codebutler: Would you like *all* fields to be copied over?

My intuition was to not expose the `api` data structures but this is a lot of duplication now.

codebutler:
Please let me know if these updates look right and I'll add unit tests. Thanks!

codebutler:
@jrbarron @ivarconr could you please take another look at this? ty!

jrbarron:
@codebutler First of all, my apologies for not getting back to this sooner. I've looked at your latest changes and they look correct from a purely technical perspective. I'm struggling a bit with the amount of duplication though and I'm sure you felt the same way when implementing it :) 

Originally the idea was to keep the API surface area of the package as small as possible which is why the API structures were never exported. However I feel the introduction of your `ListFeatures` function necessitates making those structures publicly available instead of duplicating them. The only reason I can think of for duplicating them would be if we wanted to support the internal API structures being vastly different from the public facing ones and I don't really see this happening and even if it did, it would probably warrant a new major version of the package.

So if you agree with that sentiment (?) then the next thing to decide is where to put the files. I guess there is no point in continuing to have an `internal` directory if most things are exported. Is this something you would feel comfortable trying or would you like me to come with a concrete suggestion here?

codebutler:
@ivanklee86 please take another look.

codebutler:
@jrbarron ping

codebutler:
Thanks! Who has permission to merge this in?

ivarconr:
> Thanks! Who has permission to merge this in?

@jrbarron should merge this. (He might be on holiday, it's that season.)

jrbarron:
Merged now (sorry, first day back from vacation)

dcelasun:
Hey guys, can you please tag a new release with this commit? It makes dependency management easier.

PR #73, opened 2020-10-05
Client Efficiency 

avirogalsky:
1. Set client strategies on feature initialization.
2. Update client to use strategies inside features without searching them on each `IsEnabled` call.
3. Add `EfficientStrategy` interface (enrich interface without breaking it) so it allow to load the Strategy parameters on initialization of the feature so it will not have to parse string to list on each call for example.
4. Updated `userWithIdStrategy` + test to use `EfficientStrategy` interface for example of its efficiency.


jrbarron:
@avirogalsky Is it the intention that you are targeting the `master` branch with this PR? The latest version of the client is the `v3` branch  which is the default branch for this repository. The `master` branch exists for compatibility with older versions of Go that do not support Go modules. If the intention is to target the latest branch, you probably need to rebase it on v3 and there will probably be some conflicts because some of the internal structs have now moved to being publicly exported so you also have to think about compatibility here.

avirogalsky:
> @avirogalsky Is it the intention that you are targeting the `master` branch with this PR? The latest version of the client is the `v3` branch which is the default branch for this repository. The `master` branch exists for compatibility with older versions of Go that do not support Go modules. If the intention is to target the latest branch, you probably need to rebase it on v3 and there will probably be some conflicts because some of the internal structs have now moved to being publicly exported so you also have to think about compatibility here.

@jrbarron my mistake, changed the target to v3

PR #74, opened 2020-10-07
Efficient Client v3

avirogalsky:
Hi Guys,
I've sent the PR for go client efficiency
There are some efficiency changes with full backward compatibility:
1. Move client strategies from client to repository (as repository is the one that initialises the features).
2. For each feature initialise the client's supported strategy .
3. Change IsEnabled in client so it will not search for strategies but just will call already initialised strategies in feature.
4. Add EfficientStrategy interface that allows to make strategies more efficient (only for those that implement that interface.
5. Example of using EfficientStrategy interface in userWithIdStrategy - so the string of user IDs is parsed only once to map[string]bool and each call for IsEnabled just checks existance in the map.
6. Updated tests for  userWithIdStrategy


FredrikOseberg:
@avirogalsky

I'm going to go ahead and apologize on behalf of our entire team for the way we handled this PR. For the past year we have been struggling to keep up with the community across our many repositories with a small team. 

Now that we are growing our team we are also putting in place more systemic measures to record and respond more promptly to PRs and issues. I promise we will be more vigilant in the future. 

I will close this for now, since it's so out of date with the current repository, but I thank you for your efforts and sincerely hope you'd be willing to contribute again in the future.

PR #77, opened 2020-12-06
Implement variants

ClaireLee22:
Hi guys,
This pull request is to implement variants which allow to set up multiple variants for feature toggles.
Feel free to leave comment for any improvements.
Thanks.

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -51,6 +51,18 @@ func (fr FeatureResponse) FeatureMap() map[string]interface{} {
 	return features
 }
 
+func (f Feature) GetVariant(ctx *context.Context) *Variant {
+	if f.Enabled && len(f.Variants) > 0 {
+		variant := f.getOverrideVariant(ctx)
+		if variant == nil {
+			variant = f.getVariantFromWeights(ctx)
+		}
+		variant.Enabled = true
+		return variant
+	}
+	return DISABLED_VARIANT
+}
+
 func (f Feature) getVariantFromWeights(ctx *context.Context) *Variant {
 	if len(f.Variants) > 0 {
 		totalWeight := 0
@@ -73,6 +85,17 @@ func (f Feature) getVariantFromWeights(ctx *context.Context) *Variant {
 		}
 	}
 	return DISABLED_VARIANT
+}
+
+func (f Feature) getOverrideVariant(ctx *context.Context) *Variant {
+	for _, variant := range f.Variants {
+		for _, override := range variant.Overrides {
+			if override.matchValue(ctx) {
+				return &variant
+			}
+		}
+	}
+	return nil
 }
 
 func getSeed(ctx *context.Context) string {
--- a/api/feature.go
+++ b/api/feature.go
@@ -40,7 +40,7 @@ type Feature struct {
 	Parameters ParameterMap `json:"parameters"`
 
 	// Variants is a list of variants of the feature toggle.
-	Variants []Variant `json:"variants"`
+	Variants []VariantInternal `json:"variants"`
 }
 
 func (fr FeatureResponse) FeatureMap() map[string]interface{} {
@@ -54,9 +54,12 @@ func (fr FeatureResponse) FeatureMap() map[string]interface{} {
 // Get variant for a given feature which is considered as enabled
 func (f Feature) GetVariant(ctx *context.Context) *Variant {
 	if f.Enabled && len(f.Variants) > 0 {
-		variant := f.getOverrideVariant(ctx)
-		if variant == nil {
+		v := f.getOverrideVariant(ctx)
+		var variant *Variant
+		if v == nil {
 			variant = f.getVariantFromWeights(ctx)
+		} else {
+			variant = &v.Variant
 		}
 		variant.Enabled = true
 		return variant
@@ -79,13 +82,13 @@ func (f Feature) getVariantFromWeights(ctx *context.Context) *Variant {
 		counter += uint32(variant.Weight)
 
 		if counter >= target {
-			return &variant
+			return &variant.Variant
 		}
 	}
 	return DISABLED_VARIANT
 }
 
-func (f Feature) getOverrideVariant(ctx *context.Context) *Variant {
+func (f Feature) getOverrideVariant(ctx *context.Context) *VariantInternal {
 	for _, variant := range f.Variants {
 		for _, override := range variant.Overrides {
 			if override.matchValue(ctx) {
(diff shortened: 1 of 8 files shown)


ivarconr:
Cool thanks. :fire:

ClaireLee22:
> Cool thanks. 🔥

jrbarron:
Hi @ClaireLee22 and thanks very much for your contribution! I checked out your branch and this is indeed a good first step. It seems like you have implemented the API layer, but there are still several pieces missing before variants is fully functioning. At a first glance, some of the parts that are missing are:
 * The `GetVariants` function on the client
 * Updating the `spec_test.go` to include variants
 * Unit tests
 * Reporting metrics for variants to the server

How do you want to proceed here?

ClaireLee22:
Hi @jrbarron,
Thanks for your code review and feedback.
It is really helpful for me to improve the code regarding variants.
I will keep working on that based on your instructions.
Appreciate.

ivarconr:
Really appreciate your effort here @ClaireLee22, will be good to have the Go SDK also supporting variants. Let me know if I can assist you on protocol details.

ClaireLee22:
Hi @jrbarron,
I have add variant unit test, sepc_test , GetVariant function and report variant metrics to server according to the feedback of previous code review.
Please have a look as available and feel free to let me know if anything should be improved.
Your feedback is always highly appreciated.
Thanks.

api/feature.go
```diff
@@ -41,3 +50,66 @@ func (fr FeatureResponse) FeatureMap() map[string]interface{} {
 	}
 	return features
 }
+
+func (f Feature) GetVariant(ctx *context.Context) *Variant {
```
jrbarron, line 54: This function is exported so it should be documented.

api/variant.go
```diff
@@ -0,0 +1,64 @@
+package api
+
+import "github.com/Unleash/unleash-client-go/v3/context"
+
+var DISABLED_VARIANT = &Variant{
+	Name:    "disabled",
+	Enabled: false,
+}
+
+type Payload struct {
```
jrbarron, line 10: Exported types should be documented.

api/variant.go
```diff
@@ -0,0 +1,64 @@
...
+	Type  string `json:"type"`
+	Value string `json:"value"`
+}
+
+type Override struct {
```
jrbarron, line 15: Exported types should be documented.

api/variant.go
```diff
@@ -0,0 +1,64 @@
...
+	ContextName string   `json:"contextName"`
+	Values      []string `json:"values"`
+}
+
+type Variant struct {
```
jrbarron, line 20: Exported types should be documented.

api/variant.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	for _, value := range o.Values {
+		if value == o.getIdentifier(ctx) {
+			return true
+		}
+	}
+	return false
+}
+
+func GetDefaultVariant() *Variant {
```
jrbarron, line 62: Exported functions should be documented.

jrbarron (commented):
@ClaireLee22 Great progress! I feel we are getting close now 🙂 

I checked out your code locally and ran it through the debugger. I noticed that the code implementing the client specification was not working correctly so I added some suggestions on how that can be fixed including a gist showing how I solved it.

After the above changes, the client specification executes correctly, but unfortunately most of the tests fail. Part of the problem is that the `expectedResult` in the specification file does not include the `weight` property. I will raise this issue with the Unleash team and see if they are willing to fix this (I have a patch I can send them). Unfortunately this was not the only problem, there also appears to be a bug when choosing the variant for a given user which causes some of the tests to fail. I did not have time to look more into this, but perhaps you can figure it out.

Keep up the great work! 💪

client.go
```diff
@@ -285,6 +285,36 @@ func (uc *Client) IsEnabled(feature string, options ...FeatureOption) (enabled b
 	return false
 }
 
+// GetVariant queries a variant as the specified feature is enabled.
+//
+// It is safe to call this method from multiple goroutines concurrently.
+func (uc *Client) GetVariant(feature string, fallbackVariant *api.Variant) *api.Variant {
```
jrbarron, line 291: I don't think we should force the user to pass in the second argument here (even if it can be `nil`). To keep it consistent with the `IsEnabled` function I think you should change the signature to take a variadic option type like this:

```suggestion
func (uc *Client) GetVariant(feature string, options ...VariantOption) *api.Variant {
```
This also allows us to extend the API in the future without the need to add more arguments. The `VariantOption` type would be similar to the `FeatureOption` type which is here:

https://github.com/Unleash/unleash-client-go/blob/v3/config.go#L142

If you are interested in learning more about this pattern, Dave Cheney has a great talk here:

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

jrbarron (commented):
@ClaireLee22 Great work again! I just tested this with my changes to the client spec (added weight) and all the tests are green. I have two changes which I think are worth making, but then if everything still works, I think we are ready to merge :)

FYI, I have now submitted a PR to the client specification repo to add the weights:

https://github.com/Unleash/client-specification/pull/9

jrbarron:
@ClaireLee22 I spoke with the Unleash team and it looks like the other clients do not expose "weight" and "overrides" in the public API so my PR would break the other clients. I did some playing around and managed to achieve a similar behavior to the other clients by using Go's struct embedding feature. This seems to get the tests to pass for me:

https://gist.github.com/jrbarron/4bb697ca3b66c10e325430d2de422393

Let me know what you think or if you have another idea 😉

ClaireLee22:
Hi @jrbarron,
Thanks for above great suggestions. I have revised my code again based on those feedback.

First, I believe that it is a great idea to add VariantInternal struct to embed a common variant struct and include extra weight and overrides field. I have followed your changes, and I can pass all the spec test for sure.

Secondly, as for getSeed function, I have reverted to previous function version which will generate random number as all the context value is not matched.  I might try to debug why spec test failed previously and changed that orders. However, I didn't pay attention to and forget to recover it. Thanks for the reminder.

Thirdly, I changed the GetVariant signature and add variantOption. I follow the featureOption structure and add WithVariantFallback and WithVariantFallbackFunc functions. But I am not quite sure whether if it is necessary to write those function ahead. Feel free to let me know if you have different thought or any suggestion.

Please have a look as you are available. Appreciate.

jrbarron (approved):
@ClaireLee22 Looks good and all tests are passing! 🎉 

Thanks again for your contribution. The Go Unleash community has been waiting for this one for a while 😉

ivarconr:
wow this is so amazing @ClaireLee22 :fire:

PR #78, opened 2021-01-11
fix: build with go 1.13 and 1.15

ivarconr:


jrbarron (approved):

jrbarron:
The build seems to be failing on Go 1.15. It seems like this might be a result of additional runtime checks around pointer arithmetic that were introduced in Go 1.14. There is another implementation of the murmur3 hash that apparently uses safe code at a slight cost to performance here specifically to avoid these problems with pointer arithmetic:

https://github.com/twmb/murmur3/#safety

I have pushed a commit where we switch to this other implementation.

Issue #82, opened 2021-04-21
Features read from file cannot be cast to Feature


pschmidt88:
Hello,

I think I found a bug in the current feature storage and repository implementation:
When features are read from the backup/cache file and were not (yet) fetched from the server, every feature evaluates to false/disabled.

This is because, when features are read from file, they are decoded to a generic `map[string]interface{}` structure, 
because the type of the data holding variable is a `map[string]interface{}` and the decoder doesn't know to which type it should 
decode the json to.

https://github.com/Unleash/unleash-client-go/blob/108a79f25501900055cfba97a1ae37e5b9515d9f/storage.go#L37-L41

https://github.com/Unleash/unleash-client-go/blob/108a79f25501900055cfba97a1ae37e5b9515d9f/storage.go#L58-L68

The repository then tries to cast this generic structure to a `Feature` struct (L146), which does not work, eventually returning a `nil` value although a valid feature exists.

https://github.com/Unleash/unleash-client-go/blob/108a79f25501900055cfba97a1ae37e5b9515d9f/repository.go#L141-L151

To reproduce this behavior, you can:

1. Create a client with a custom backup path
2. Let the client persist features fetched from the server into the schema json file
3. Stop the server
4. Restart the client and let it read the features from the file
5. Query an enabled toggle

The feature toggle will evaluate to false

pschmidt88:
Is there a reason why `data` in `defaultStorage` is of type `map[string]interface{}` instead of `map[string]api.Feature` ?

pschmidt88:
This was fixed in #83

PR #83, opened 2021-04-23
Parse features from schema file into Feature struct

pschmidt88:
This is a possible fix for issue #82

The features read from file will be parsed into a `map[string]api.Feature`, then
copied over to the `data` property of the `defaultStorage` struct.
This way they can be cast correctly to a feature when read in `repository.getToggle()`.

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/Unleash/unleash-client-go/v3/api"
 )
 
 // Storage is an interface that can be implemented in order to have control over how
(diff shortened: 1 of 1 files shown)


jrbarron (approved):
@pschmidt88 Thanks for your contribution! This seems like a reasonable fix. Ideally the `Storage` interface would be updated to use `map[string]api.Feature` as well, but I think that will have to wait until the next major release for compatibility reasons.

As to why this is like this, I can't really remember, but the original version of the library did not export the types from the `api` package so I suspect that is the reason. I'll wait for feedback from @FredrikOseberg before merging, but this seems like a reasonable fix for now. A unit test would perhaps be nice, but there wasn't one there to begin with so can't really fault you there 😆

FredrikOseberg (approved):
LGTM, thanks for this fix.

PR #86, opened 2021-06-04
Expose defaultStorage

kskumgk63:
I would like to embed defaultStorage to my original struct because storage customization would be easy.

Related slack link: https://unleash-community.slack.com/archives/CGN5JT69F/p1622789881074700

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -42,7 +42,7 @@ func newRepository(options repositoryOptions, channels repositoryChannels) *repo
 	}
 
 	if options.storage == nil {
-		repo.options.storage = &defaultStorage{}
+		repo.options.storage = &DefaultStorage{}
 	}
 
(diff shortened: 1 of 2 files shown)


chriswk (approved):
You could potentially break the client with too much modification. However, I think this seems like a reasonable change

chriswk:
Hi @jrbarron - Talked to @FredrikOseberg  on this and we couldn't think of a major argument for not exposing this, but I'd love to have your opinion on this as well before merging.

jrbarron:
Seems reasonable to me :) It was most likely private to reduce the API surface area of the package and since it was the first implementation of the storage engine we did not know how long it would survive at that point.

jrbarron (approved):

kskumgk63:
@chriswk @jrbarron 
Thanks for approves, Could you merge this PR?

PR #88, opened 2021-06-23
fix: add override to GetVariant

FredrikOseberg:
Fixes #88


chriswk (changes requested):
I'm missing a `WithContext(*context.Context) VariantOption` method to be more in line with how we add VariantFallback and/or VariantFallbackFunc

chriswk (approved):
Actually this seems fine.

PR #89, opened 2021-07-02
Fix/variant bugfix

FredrikOseberg:
- Check isEnabled before returning variants
- Initalize variant map in feature metrics bucket


ivarconr (approved):
Nice job 👍

chriswk (approved):

PR #97, opened 2022-02-24
feat: add new constraint operators

olav:


internal/constraints/operator_num.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func operatorNum(
+	ctx *context.Context,
+	constraint api.Constraint,
+	check func(context *big.Float, constraint *big.Float) bool,
+) (bool, error) {
+	contextValue := ctx.Field(constraint.ContextName)
+	contextParsed, _, contextErr := big.ParseFloat(contextValue, 10, 0, big.ToNearestEven)
+	constraintParsed, _, constraintErr := big.ParseFloat(constraint.Value, 10, 0, big.ToNearestEven)
```
chriswk, line 46: This one is interesting, I know we in rust did equal as absolute difference between a and b less than :epsilon: which for rust's f64 is `2.2204460492503131*10^-16` or `2^-52` - as in, rust does not do rounding. Neither does javascript, which treats it as a Number which also does equivalence as `absdiff < Number.EPSILON`.

chriswk: For current client specifications which only tests that the sdk can handle floats, not how they handle difference this is ok.

chriswk (approved):
Great job. The question about how we should handle numeric difference for floats is not directly related to this PR, so am not going to let it prevent me from approving.

It is a discussion we should have though, so that our SDKs have consistent behaviour.

internal/constraints/operator_date_test.go
```diff
@@ -0,0 +1,102 @@
...
+			expected:    false,
+		},
+		{
+			ctx:         &context.Context{Properties: map[string]string{"date": "2006-01-02T15:04:05.999999998Z"}},
+			constraints: []api.Constraint{{ContextName: "date", Operator: "DATE_AFTER", Value: "2006-01-02T15:04:05.999999999Z"}},
+			expected:    false,
+		},
+		{
+			ctx:         &context.Context{Properties: map[string]string{"date": "2006-01-02T15:04:05.999999999Z"}},
+			constraints: []api.Constraint{{ContextName: "date", Operator: "DATE_AFTER", Value: "2006-01-02T15:04:05.999999998Z"}},
```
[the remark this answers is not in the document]
olav: Yeah, I was mostly wondering if it would work out of the box, looks like it does! Can probably remove these though.

internal/constraints/operator_num.go
olav: Yeah, wasn't sure about this one. This seemed like a sensible, built-in way to compare floats, but would be nice with a common approach.

chriswk:
[Rust SDK: add support for new strategy constraint operators](https://trello.com/c/i4FBjtOT/895-rust-sdk-add-support-for-new-strategy-constraint-operators)

Issue #81, opened 2021-03-25
feat: Custom stickiness for variants


ivarconr:
This is kind of similar to #80 , but for variants. 

The stickiness will be defined as a field on all the variants in the variants array, this to ensure we are not breaking our protocol. In the example we can see the stickiness is configured to "tenantId". 

```json
{"variants": [
     {
		"name": "yellow",
		"weight": 166,
		"weightType": "variable",
		"payload": {
			"type": "string",
			"value": "yellow submarine"
		},
		"overrides": [],
		"stickiness": "tenantId"
	}, {
		"name": "salmon",
		"weight": 166,
		"weightType": "variable",
		"payload": {
			"type": "string",
			"value": "salmon"
		},
		"overrides": [],
		"stickiness": "tenantId"
	}
]}
```

How it looks in Unleash Admin UI:
![image](https://user-images.githubusercontent.com/158948/112531447-deaf0900-8da7-11eb-8995-d8dd56b5d374.png)

**Edge cases:**
If no stickiness is defined the SDK should assume "default" stickiness as today. 

**To guide the implementation we have:**
- added new client-specifications, see [pr-11](https://github.com/Unleash/client-specification/pull/11)
- implemented PoC in Node SDK [see pr-202](https://github.com/Unleash/unleash-client-node/pull/202/files)

Issue #80, opened 2021-03-25
feat: FlexibleRollout should support "custom" stickiness.


ivarconr:
The [Flexible Rollout strategy](https://docs.getunleash.io/docs/user_guide/activation_strategy#flexiblerollout) supports multiple options for `stickiness`, used to define how we guarantee consistency for a gradual rollout.

Today it support the following options:
- `default` (Unleash chooses the first value present on the context in defined order userId, sessionId, random.)
- `userId`
- `sessionId`
- `random`

We have now extended the protocol to support any field in the [unleash context](https://docs.getunleash.io/docs/user_guide/unleash_context). In addition  It can be any of the predefined context field, in addition to any custom properties the user has defined. 

This means that the "stickiness" parameter in the strategy configuration can now be any field on the context, and the SDK should use the specified context field to calculate the "hash" which determine whether the activation strategy should evaluate to true or false. 

How it looks in the UI:
![image](https://user-images.githubusercontent.com/158948/112530370-a65afb00-8da6-11eb-9874-1dbf876838be.png)

Edge case:
- If the specified context field is not specific the activation strategy should evaluate to false. 

To guide the implementation we have:
- added new client-specifications, see [pr-11](https://github.com/Unleash/client-specification/pull/11)
- implemented PoC in Node SDK [see pr-201](https://github.com/Unleash/unleash-client-node/pull/201/files)

PR #99, opened 2022-03-08
feat: Implement custom stickiness for FlexibleRollout and variants

kjvalencik:
There is some duplication in field matching and coalescing, but I chose not to address it in this PR this since the duplication was existing.

I chose to implement both features in a single PR since the changes were fairly small. Implementing in a single PR allows enabling the full spec test.

Implements https://github.com/Unleash/unleash-client-go/issues/80
Implements https://github.com/Unleash/unleash-client-go/issues/81

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -76,7 +76,9 @@ func (f Feature) getVariantFromWeights(ctx *context.Context) *Variant {
 		return DISABLED_VARIANT
 	}
 
-	target := getNormalizedNumber(getSeed(ctx), f.Name, totalWeight)
+	stickiness := f.Variants[0].Stickiness
+
+	target := getNormalizedNumber(getSeed(ctx, stickiness), f.Name, totalWeight)
 	counter := uint32(0)
 	for _, variant := range f.Variants {
 		counter += uint32(variant.Weight)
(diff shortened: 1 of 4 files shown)


api/feature.go
```diff
@@ -76,7 +76,9 @@ func (f Feature) getVariantFromWeights(ctx *context.Context) *Variant {
 		return DISABLED_VARIANT
 	}
 
-	target := getNormalizedNumber(getSeed(ctx), f.Name, totalWeight)
+	stickiness := f.Variants[0].Stickiness
```
kjvalencik, line 79: This matches the [node client implementation](https://github.com/Unleash/unleash-client-node/blob/dda0aa9d75e2d979002d87b8f5b8f08e5dd91532/src/variant.ts#L88). The array is guaranteed to have a length of at least `1`, otherwise `totalWeight` would be `0` and the function would have already returned.

chriswk:
Superseded by #107 . We had to rebase on top of main and handle that validation tests now could error with invalid data. The logic for custom stickiness is still the same. 

We're sorry for the long turn around here, but we deeply appreciate you taking the time to make this PR :)

PR #102, opened 2022-03-10
fix: empty userWithIdStrategy case

icecrime:
I'm new to Unleash, so my apologies if this pull request turns out to be nonsensical.

`IsEnabled` returns true for an empty UserIdStrategy when `UserId` is missing from the context. The reason is that `strings.Split` for an empty string returns `[""]`, giving the `userId` comparison loop a chance to both run _and_ find a match.


ivarconr (approved):

chriswk (approved):
Looks good to me. Nice catch @icecrime

PR #108, opened 2022-04-28
docs: mention default intervals

chriswk:
It's currently not obvious from the documentation if they should call `WithInterval`/`WithMetricsInterval` or if the default values are appropriate for the consumer. Mentioning these defaults should make this a bit more straightforward.

This is the work of @costela, but I managed to merge it to master instead of v3, this just moves the commit


FredrikOseberg (approved):
LGTM

PR #116, opened 2022-07-06
Feat/global segments

### LINKED ISSUE #115: Implement Global Segments

Just a feature that needs to be implemented - Global Segments. Unleash v4.13 supports enhanced responses for global segments, it would be great if this SDK can make use of this.

### Background

Segments are effectively a way for Unleash users to define a list of constraints in such a way that makes them reusable across toggles without manually copying the constraint across to another toggle. Segments have two modes of operation, from the SDK's perspective, the inline mode will have no impact, segments will be remapped on the server side into constraints on the toggle information, no changes need to be . The second mode, global segments, requires that the SDK both opt in and handle the response differently. The handling should effectively result in unpacking the segments referenced in the feature strategies into a set of constraints. The changes required are described below.

### Solution suggestions

**Control Header**

The SDK needs to pass up a `Unleash-Client-Spec` header with a semver value greater or equal to `4.2.0`  (i.e. be greater or equal to the version of the [unleash client spec tests](https://github.com/Unleash/client-specification) where global segments are described) when hitting the get toggles endpoint on the Unleash server. This will enable the Unleash server to respond with the enhanced format

Example of the difference between enhanced and standard format:

**Standard Format (default)**

```
{
   "version": 2,
   "features": [  
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [
                       {
                           "values": [
                               "31"
                           ],
                           "inverted": false,
                           "operator": "IN",
                           "contextName": "appName",
                           "caseInsensitive": false
                       }
                   ],          
                   "parameters": {
                       "groupId": "Test1",
                       "rollout": "100",
                       "stickiness": "default"
                   }
               }
           ],
           "name": "Test1"
       },
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [
                       {
                           "values": [
                               "31"
                           ],
                           "inverted": false,
                           "operator": "IN",
                           "contextName": "appName",
                           "caseInsensitive": false
                       }
                   ],          
                   "parameters": {
                       "groupId": "Test2",
                       "rollout": "100",
                       "stickiness": "default"
                   }
               }
           ],
           "name": "Test2"
       }    
   ],
   "query": {
       "environment": "default"
   }
}

```

***Enhanced Format (requires opt in)***
 
```
{
   "version": 2,
   "features": [   
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [],
                   "parameters": {
                       "groupId": "Test1",
                       "rollout": "100",
                       "stickiness": "default"
                   },
                   "segments": [
                       1
                   ]
               }
           ],
           "name": "Test1"
       },
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [],
                   "parameters": {
                       "groupId": "Test2",
                       "rollout": "100",
                       "stickiness": "default"
                   },
                   "segments": [
                       1
                   ]
               }
           ],
           "name": "Test2"
       }     
   ],
   "query": {
       "environment": "default"
   },
   "segments": [
       {
           "id": 1,           
           "constraints": [
               {
                   "values": [
                       "31"
                   ],
                   "inverted": false,
                   "operator": "IN",
                   "contextName": "appName",
                   "caseInsensitive": false
               }
           ]           
       }
   ]
}
```
The relevant changes between the two formats are that in the enhanced format the segments are defined once as a global list and referenced within the strategy on the toggle by its ID. What's important to note is that the two above packets should be 
handled identically, they reference the same toggle state.

**Considerations**

- Global segments are intended to handle large datasets, how large has not been formally specified yet but expectations are around 1 000 to 10 000 segments across 1 000 toggles. As a result, time and space complexity of the implementations needs to be considered. 
- In the case of global segments, if the mapping from segment id to segment is incomplete i.e. a segment id is referenced in a toggle strategy that doesn’t map back to the global segments list, then the toggle should be evaluated to false. This is enforced through one of the client specification tests in v4.2.0 of the client spec
- A reference implementation is provided in node JS: https://github.com/Unleash/unleash-client-node/pull/329 (note that this doesn't include the header, that can be seen here: https://github.com/Unleash/unleash-client-node/pull/335)

FredrikOseberg:
This PR will implement global segments as a part of performance optimizations related to segment use. This will ensure that segments are sent as part of the payload from the client API, parsed correctly into a map on the SDK side and combined with the regular strategy constraints.

Fixes #115

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Unleash/unleash-client-go/v3/api"
 )
 
+var SEGMENT_CLIENT_SPEC_VERSION = "4.2.0"
 type repository struct {
 	repositoryChannels
 	sync.RWMutex
@@ -95,6 +96,9 @@ func (r *repository) fetch() error {
 	req.Header.Add("UNLEASH-APPNAME", r.options.appName)
 	req.Header.Add("UNLEASH-INSTANCEID", r.options.instanceId)
 	req.Header.Add("User-Agent", r.options.appName)
+	// Needs to reference a version of the client specifications that include
+	// global segments
+	req.Header.Add("Unleash-Client-Spec", SEGMENT_CLIENT_SPEC_VERSION)
 
 	for k, v := range r.options.customHeaders {
 		req.Header[k] = v
@@ -153,16 +157,18 @@ func (r *repository) getToggle(key string) *api.Feature {
 	return nil
 }
 
-func (r *repository) resolveSegmentConstraints(toggleName string) []api.Constraint {
-	toggle := r.getToggle(toggleName);
+func (r *repository) resolveSegmentConstraints(strategy api.Strategy) (error, []api.Constraint) {
 	segmentConstraints := []api.Constraint{}
 
-	for _, segmentId := range toggle.Segments {
-		resolvedConstraints := r.segments[segmentId];
-		segmentConstraints = append(segmentConstraints, resolvedConstraints...)
+	for _, segmentId := range strategy.Segments {
+		if resolvedConstraints, ok := r.segments[segmentId]; ok {
+			segmentConstraints = append(segmentConstraints, resolvedConstraints...)
+		} else {
+			return fmt.Errorf("segment does not exists"), segmentConstraints
+		}
 	}
 
-	return segmentConstraints;
+	return nil, segmentConstraints;
 }
 
 func (r *repository) list() []api.Feature {
--- a/repository.go
+++ b/repository.go
@@ -13,6 +13,7 @@ import (
 )
 
 var SEGMENT_CLIENT_SPEC_VERSION = "4.2.0"
+
 type repository struct {
 	repositoryChannels
 	sync.RWMutex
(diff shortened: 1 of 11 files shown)


repository.go
```diff
@@ -150,6 +158,20 @@ func (r *repository) getToggle(key string) *api.Feature {
...
 }
 
+func (r *repository) resolveSegmentConstraints(strategy api.Strategy) ([]api.Constraint, error) {
+	segmentConstraints := []api.Constraint{}
+
+	for _, segmentId := range strategy.Segments {
+		if resolvedConstraints, ok := r.segments[segmentId]; ok {
+			segmentConstraints = append(segmentConstraints, resolvedConstraints...)
+		} else {
+			return segmentConstraints, fmt.Errorf("segment does not exists")
```
sighphyre:
```suggestion
			return segmentConstraints, fmt.Errorf("segment does not exist")
```

sighphyre (approved):
Looks good to me! I've left a text change comment, aside from that, this is great!

Just a quick double check because my Golang isn't strong - this should still parse responses from the API when the segments property isn't set?

Edit: Just saw this work with empty responses, great!

FredrikOseberg:
> Looks good to me! I've left a text change comment, aside from that, this is great!
> 
> Just a quick double check because my Golang isn't strong - this should still parse responses from the API when the segments property isn't set?
> 
> Edit: Just saw this work with empty responses, great!

I added a test for testing the old response with the new interface: 

https://github.com/Unleash/unleash-client-go/pull/116/commits/243af78842d9e22720a27e512ab779cc2a879e20

PR #119, opened 2022-07-23
#118 Add IsFeatureEnabled function to client

### LINKED ISSUE #118: (Feature Request) Enable passing api.Feature instead of a string to an alternative of IsEnabled

I would like a way to call a variant of IsEnabled that can accept an api.Feature without needing to pull it from the `uc.repository.getToggle(feature)` .

### Background

I have been using this client in a way where I have the api.Feature that needs to be checked and having it retrieved from the repository is adding a few milliseconds of unnecessary latency.

### Solution suggestions

I'm willing to add a variant function of `IsEnabled ` that accepts ` api.Feature` as the first parameter instead of a string.

Advantage:
- Enable more flexibility or control when interfacing with the client.
- More performant use of the `IsEnabled` function for those with access to the `api.Feature` from previous processing.

Drawbacks:
- Would add another public function to the client that could confuse people. (I feel like this is a stretch of a drawback)

courupteddata:
## About the changes
This change adds an IsFeatureEnabled function that is a variant of the IsEnabled function but accepts a api.Feature instead of a string for the case where you already have the feature.

Closes #118

## Discussion points
I'm not certain on the function name and don't want to cause confusion, but I thought it made sense.

I think this functionality could make sense for others and if there is a better approach someone else wants I'm all ears. 👂

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,47 @@ unleash.IsEnabled("someToggle", unleash.WithContext(ctx))
 
 This client uses go routines to report several events and doesn't drain the channel by default. So you need to either register a listener using `WithListener` or drain the channel "manually" (demonstrated in [this example](https://github.com/Unleash/unleash-client-go/blob/master/example_with_instance_test.go)).
 
+### Feature Resolver
+
+`FeatureResolver` is a `FeatureOption` used in `IsEnabled` via the `WithResolver`. 
+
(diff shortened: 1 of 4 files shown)


FredrikOseberg (commented):
@courupteddata Could you add a section to the readme in this PR on how to use this new functionality?

courupteddata:
@FredrikOseberg I've added a readme entry for this change.

FredrikOseberg (approved):
LGTM. Thanks for your contribution and patience @courupteddata

PR #120, opened 2022-11-10
Adding option to query for environment

ljupcho:
This PR fix will add the environment variable to the http request when querying for feature for endpoint `api/client/features`. 
At the moment the endpoint  would only query for the project name. With passing in the environment to the query string i am able to toggle IsEnabled true|false for a feature for development environment.

### DIFF
--- a/client.go
+++ b/client.go
@@ -159,6 +159,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			url:             *parsedUrl,
 			appName:         uc.options.appName,
 			projectName:     uc.options.projectName,
+			environment:     uc.options.environment,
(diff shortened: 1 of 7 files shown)


FredrikOseberg (commented):
Hi @ljupcho 

I'm not sure this change makes sense. Currently the only way to work with environments in unleash is through the API tokens. The environment is coded into the API token when you create it. You can read more about it here: https://docs.getunleash.io/reference/api-tokens-and-client-keys#token-data

As such, the source of truth for what kind of configuration the SDK will retrieve is based on the token you use with the SDK, not the query parameter.

I will close this for now, please let me know if I missed something.

PR #121, opened 2022-11-28
fix: avoid panic when not initialized

costela:
## About the changes
Using `IsEnabled` withoug initialization currently panics with a nil deref.

This change eases e.g. integration of unleash in frameworks, allowing optional initialization.

Consumers already have to deal with the fact that the unleash server may be unreachable during a deployment, so even if the initialization works, to be defensive one must assume a _disabled_ feature flag is less harmful than a panicking service.

### DIFF
--- a/unleash.go
+++ b/unleash.go
@@ -37,6 +37,9 @@ type RepositoryListener interface {
 
 // IsEnabled queries the default client whether or not the specified feature is enabled or not.
 func IsEnabled(feature string, options ...FeatureOption) bool {
+	if defaultClient == nil {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


FredrikOseberg (approved):
Seems like a reasonable change! LGTM.

I wonder if we should return the default variant from the GetVariant call if there is no defaultClient to be consistent with the calls here.

costela:
@FredrikOseberg yes, very good point! I took the liberty of also adding a fallback to `Close()`, although I suspect this is less relevant.

FredrikOseberg:
Thanks! I will cut a release including these changes.

PR #123, opened 2022-12-12
force getVariant to also add to the feature toggle metrics counter

sighphyre:
This forces resolving a variant to correctly add to the feature toggle counter metrics. Also fixes an issue where variants wouldn't get counted for the default/disabled case (brings this SDK inline with the behaviour of its siblings)

### DIFF
--- a/client.go
+++ b/client.go
@@ -315,6 +315,15 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled b
 //
 // It is safe to call this method from multiple goroutines concurrently.
 func (uc *Client) GetVariant(feature string, options ...VariantOption) *api.Variant {
+	variant, enabled := uc.getVariant(feature, options...)
+	defer func() {
+		uc.metrics.countVariants(feature, enabled, variant.Name)
+	}()
+	return variant
+}
+
+// getVariant abstracts away the logic for resolving a variant without metrics
+func (uc *Client) getVariant(feature string, options ...VariantOption) (*api.Variant, bool) {
 	defaultVariant := api.GetDefaultVariant()
 	var opts variantOption
 	for _, o := range options {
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -315,6 +315,15 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled b
...
 func (uc *Client) GetVariant(feature string, options ...VariantOption) *api.Variant {
+	variant, enabled := uc.getVariant(feature, options...)
+	defer func() {
+		uc.metrics.countVariants(feature, enabled, variant.Name)
+	}()
+	return variant
+}
+
+// getVariant abstracts away the logic for resolving a variant without metrics
+func (uc *Client) getVariant(feature string, options ...VariantOption) (*api.Variant, bool) {
```
sighphyre, line 326: Moves the variant resolution code into its own method so that we can correctly count disabled/default variants

metrics.go
```diff
@@ -230,11 +230,13 @@ func (m *metrics) count(name string, enabled bool) {
...
 }
 
-func (m *metrics) countVariants(name string, variantName string) {
+func (m *metrics) countVariants(name string, enabled bool, variantName string) {
 	if m.options.disableMetrics {
 		return
 	}
 
+	m.add(name, enabled, 1)
+	m.metricsChannels.count <- metric{Name: name, Enabled: enabled}
```
sighphyre, line 239: I believe this is necessary but I'm not 100% sure - since we're now incrementing the counter for feature toggles, it makes sense to now add this to the channel

gastonfournier:
I have refactored in a way I think it's simpler to maintain: https://github.com/Unleash/unleash-client-go/pull/124 but having issues with the coverage report which I can't make sense of

gastonfournier:
Closing in favor of https://github.com/Unleash/unleash-client-go/pull/124

PR #124, opened 2022-12-15
force getVariant to also add to the feature toggle metrics counter

gastonfournier:
This forces resolving a variant to correctly add to the feature toggle counter metrics. Also fixes an issue where variants wouldn't get counted for the default/disabled case (brings this SDK inline with the behaviour of its siblings)

Also, other unrelated changes:
1. Moved unleash_test.go to unleash_mock.go because it was in `unleash` package and not having any test
2. Improved the Readme due to an unexisting command


metrics.go
```diff
@@ -230,11 +230,13 @@ func (m *metrics) count(name string, enabled bool) {
...
 }
 
-func (m *metrics) countVariants(name string, variantName string) {
+func (m *metrics) countVariants(name string, enabled bool, variantName string) {
 	if m.options.disableMetrics {
 		return
 	}
 
+	m.add(name, enabled, 1)
+	m.metricsChannels.count <- metric{Name: name, Enabled: enabled}
```
gastonfournier, line 239: I believe this is necessary but I'm not 100% sure - since we're now incrementing the counter for feature toggles, it makes sense to now add this to the channel

Same as https://github.com/Unleash/unleash-client-go/pull/123#discussion_r1045534318

example_with_variants_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if err != nil {
+		t.Fail()
+	}
+
+	enabled := unleash.GetVariant("ActivityLog")
+	fmt.Printf("feature is enabled? %v\n", enabled.Enabled)
+	if enabled.Enabled == true {
+		t.Fatalf("Expected feature to be enabled")
+	}
```
gastonfournier: I believe that according to the spec `enabled.Enabled` should be true, but it's actually false:
```json
{
      "name": "ActivityLog",
      "type": "release",
      "enabled": true,
      "stale": false,
      "strategies": [
        {
          "name": "Environment",
          "parameters": {
            "Dev": "",
            "QA": "",
            "Test": "",
            "Training": "",
            "Prod": ""
          },
          "constraints": []
        }
      ],
      "variants": [
        {
          "name": "type_1",
          "weight": 500,
          "payload": { "type": "string", "value": "opt_1" },
          "overrides": [],
          "weightType": "variable",
          "stickiness": "default"
        },
        {
          "name": "type_2",
          "weight": 500,
          "payload": { "type": "string", "value": "opt_2" },
          "overrides": [],
          "weightType": "variable",
          "stickiness": "default"
        }
      ]
    },
```

gastonfournier: This one was not acting as we expected because of the strategy named `Environment` that doesn't exist in the default Unleash configuration. We solved this by switching to another feature toggle using the `default` strategy

client.go
```diff
@@ -315,6 +315,15 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled b
...
 func (uc *Client) GetVariant(feature string, options ...VariantOption) *api.Variant {
+	variant := uc.GetVariantWithoutMetrics(feature, options...)
+	defer func() {
+		uc.metrics.countVariants(feature, variant.Enabled, variant.Name)
+	}()
+	return variant
+}
+
+// GetVariantWithoutMetrics abstracts away the logic for resolving a variant without metrics
+func (uc *Client) GetVariantWithoutMetrics(feature string, options ...VariantOption) *api.Variant {
```
FredrikOseberg: I think this should be lowercase, we don't want to expose this through the public API (by default uppercase method names are public in go, and lowercase are private).

gastonfournier: Thanks! I didn't know that! 

@nunogois learning opportunity here :) https://go.dev/doc/effective_go#names

gastonfournier: [9941bc6](https://github.com/Unleash/unleash-client-go/pull/124/commits/9941bc664d535df08fd8b32a28966a73c07eaee4)

chriswk (approved):
Looks good to me

PR #135, opened 2023-08-02
feat: add variant strategy support

sjaanus:


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         name: Checkout client specifications
         with:
           repository: Unleash/client-specification
-          ref: refs/tags/v4.1.0
+          ref: refs/tags/v4.3.1
```
sjaanus, line 18: Update client spec to latest

api/feature.go
```diff
@@ -63,12 +63,12 @@ func (fr FeatureResponse) SegmentsMap() map[int][]Constraint {
...
 	}
 
-	return segments;
+	return segments
 }
 
 // Get variant for a given feature which is considered as enabled
-func (f Feature) GetVariant(ctx *context.Context) *Variant {
-	if f.Enabled && len(f.Variants) > 0 {
+func (f VariantCollection) GetVariant(ctx *context.Context) *Variant {
```
sjaanus: Instead of feature, now variant can be resolved from Strategy.

client.go
```diff
@@ -243,12 +243,12 @@ func (uc *Client) IsEnabled(feature string, options ...FeatureOption) (enabled b
...
 	}()
 
-	return uc.isEnabled(feature, options...)
+	return uc.isEnabled(feature, options...).Enabled
 }
 
 // isEnabled abstracts away the details of checking if a toggle is turned on or off
 // without metrics
-func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled bool) {
+func (uc *Client) isEnabled(feature string, options ...FeatureOption) api.StrategyResult {
```
sjaanus, line 251: Instead of simple boolean, we are now returning complex object, that includes variant.

client.go
```diff
@@ -304,11 +316,33 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled b
 		if ok, err := constraints.Check(ctx, allConstraints); err != nil {
 			uc.errors <- err
 		} else if ok && foundStrategy.IsEnabled(s.Parameters, ctx) {
-			return true
+			if s.Variants != nil && len(s.Variants) > 0 {
+				groupIdValue := s.Parameters[strategy.ParamGroupId]
+				groupId, ok := groupIdValue.(string)
+				if !ok {
```
sjaanus, line 322: If strategy does not have groupId, then bad data and return disabled

client.go
```diff
@@ -304,11 +316,33 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) (enabled b
...
+
+				return api.StrategyResult{
+					Enabled: true,
+					Variant: api.VariantCollection{
+						GroupId:  groupId,
+						Variants: s.Variants,
+					}.GetVariant(ctx),
+				}
+			} else {
+				return api.StrategyResult{
```
sjaanus, line 336: If strategy does not have variants, just return true and use old variants.

client.go
```diff
@@ -365,11 +401,18 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
...
+	if strategyResult.Variant != nil {
+		return strategyResult.Variant
+	}
+
 	if len(f.Variants) == 0 {
 		return defaultVariant
 	}
 
-	return f.GetVariant(ctx)
+	return api.VariantCollection{
```
sjaanus, line 412: If code runs here, then strategy variants were not found, use old variants

kwasniew (approved):

PR #137, opened 2023-08-23
Use fallback value if client is not initialized

Lucaber:
## About the changes
If no defaultClient is initialized, we should return the provided fallback value instead of `false`.


FredrikOseberg:
> ## About the changes
> If no defaultClient is initialized, we should return the provided fallback value instead of `false`.

@Lucaber 

Apologies it took us this long to get it to this. This looks reasonable to me, would it be possible for you to fix the conflict here?

Lucaber:
rebased

FredrikOseberg (approved):
LGTM

PR #138, opened 2023-09-01
test: benchmark feature toggle evaluation

nunogois:
After discussing with @gardleopard we were interested in a naïve and simple benchmark for feature toggle evaluation in a language like Go.

This can be ran with a command like: `go test -run=^$ -bench=BenchmarkFeatureToggleEvaluation -benchtime=10s`

Here's an example of how the output could look like:
> ❯ go test -run=^$ -bench=BenchmarkFeatureToggleEvaluation -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/Unleash/unleash-client-go/v3
BenchmarkFeatureToggleEvaluation-8      Final Estimated Operations Per Day: 101.131 billion (1.011315e+11)
13635154               854.3 ns/op
PASS
ok      github.com/Unleash/unleash-client-go/v3 13.388s

The `-8` in `BenchmarkFeatureToggleEvaluation-8` stands for 8 cores (all the cores in my M1 CPU).
**Clarification**: This version of the code is actually only using 1 core, no parallelism.

I couldn't find a way of initializing our Unleash client without a listener, so I added a simple `NoOpListener`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,9 @@ Run race-tests:
 
 You can benchmark feature toggle evaluation by running:
 
-    go test -run=^$ -bench=BenchmarkFeatureToggleEvaluation -benchtime=10s
+```
+go test -run=^$ -bench=BenchmarkFeatureToggleEvaluation -benchtime=10s
+```
 
 Here's an example of how the output could look like:
 
--- a/README.md
+++ b/README.md
@@ -203,11 +203,12 @@ This client uses go routines to report several events and doesn't drain the chan
 
 ### Feature Resolver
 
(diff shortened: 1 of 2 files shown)


gardleopard (approved):
We tested this together and I love it! Would be good to add the how to run this in the README.md as well.

nunogois:
> We tested this together and I love it! Would be good to add the how to run this in the README.md as well.

Added a new mention to the benchmark in the README, let me know what you think. Addressed in https://github.com/Unleash/unleash-client-go/pull/138/commits/b9dc62449ffb3413ba9468646a112064e58c2a0c

PR #140, opened 2023-09-26
Feat/dependent feature toggles

FredrikOseberg:
Implements dependent feature toggles for the golang-sdk

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -53,11 +53,12 @@ type Feature struct {
 }
 
 type FeatureDependencies struct {
-	// The name of the feature toggle we depend upon
+	// Feature is the name of the feature toggle we depend upon
 	Feature string `json:"feature"`
-	// Possible support for depending on a variant in the future
+	// Variants contains a string of variants that the dependency should resolve to
 	Variants *[]string `json:"variants"`
-
+	// Enabled is the property that determines whether the dependency should be on or off
+	// If the property is absent from the payload it's assumed to be default on
 	Enabled *bool `json:"enabled"`
 } 	
 
--- a/api/feature.go
+++ b/api/feature.go
@@ -47,7 +47,19 @@ type Feature struct {
 
 	// Variants is a list of variants of the feature toggle.
 	Variants []VariantInternal `json:"variants"`
+
+	// Dependencies is a list of feature toggle dependency objects
+	Dependencies *[]FeatureDependencies `json:"dependencies"`
 }
+
+type FeatureDependencies struct {
+	// The name of the feature toggle we depend upon
+	Feature string `json:"feature"`
+	// Possible support for depending on a variant in the future
+	Variants *[]string `json:"variants"`
+
+	Enabled *bool `json:"enabled"`
+} 	
 
 func (fr FeatureResponse) FeatureMap() map[string]interface{} {
 	features := map[string]interface{}{}
--- a/api/feature.go
+++ b/api/feature.go
@@ -49,10 +49,10 @@ type Feature struct {
 	Variants []VariantInternal `json:"variants"`
 
 	// Dependencies is a list of feature toggle dependency objects
-	Dependencies *[]FeatureDependencies `json:"dependencies"`
+	Dependencies *[]Dependency `json:"dependencies"`
 }
 
-type FeatureDependencies struct {
+type Dependency struct {
 	// Feature is the name of the feature toggle we depend upon
 	Feature string `json:"feature"`
 	// Variants contains a string of variants that the dependency should resolve to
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -345,6 +340,48 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) api.Strate
...
+        if parentToggle.Dependencies != nil && len(*parentToggle.Dependencies) > 0 {
+            return false
+        }
+
+		// According to the schema, if the enabled property is absent we assume it's true.
+        if parent.Enabled == nil {
+            if parent.Variants != nil && len(*parent.Variants) > 0 {
+                variantName := uc.getVariantWithoutMetrics(parent.Feature, WithVariantContext(context)).Name
+                if contains(*parent.Variants, variantName) {
+                    continue
```
kwasniew: not sure how idiomatic continue is in golang but I'd rather negate the contains and early exit with return false

client.go
```diff
@@ -345,6 +340,48 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) api.Strate
 	}
 }
 
+func (uc *Client) isParentDependencySatisfied(feature *api.Feature, context context.Context) bool {
```
kwasniew, line 342:
```
func (uc *Client) isParentDependencySatisfied(feature *api.Feature, context context.Context) bool {
    if feature == nil || feature.Dependencies == nil || len(*feature.Dependencies) == 0 {
        return true
    }

    for _, parent := range *feature.Dependencies {
        parentToggle := uc.repository.getToggle(parent.Feature)
        isEnabled := uc.isEnabled(parent.Feature, WithContext(context)).Enabled

        if parentToggle == nil || 
           (parentToggle.Dependencies != nil && len(*parentToggle.Dependencies) > 0) || 
           (parent.Enabled != nil && !isEnabled) {
            return false
        }

        if parent.Enabled == nil {
            variantName := uc.getVariantWithoutMetrics(parent.Feature, WithVariantContext(context)).Name
            hasValidVariant := parent.Variants != nil && len(*parent.Variants) > 0 && contains(*parent.Variants, variantName)
            
            if !hasValidVariant && !isEnabled {
                return false
            }
        }
    }

    return true
}
```

FredrikOseberg: Unless we write an every helper, we can't return out here or we will exit the entire function without properly looping over all the dependencies.

api/feature.go
```diff
@@ -47,6 +47,19 @@ type Feature struct {
 
 	// Variants is a list of variants of the feature toggle.
 	Variants []VariantInternal `json:"variants"`
+
+	// Dependencies is a list of feature toggle dependency objects
+	Dependencies *[]FeatureDependencies `json:"dependencies"`
+}
+
+type FeatureDependencies struct {
```
kwasniew: I called this just Dependency in Node SDK with an option to other dependencies. The property inside Feature is indicating what type of dependency it is. If we don't expose this type in the public API then I'm ok with any name.

client.go
```diff
@@ -254,31 +254,25 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) api.Strate
...
-		if opts.fallbackFunc != nil {
-			return api.StrategyResult{
-				Enabled: opts.fallbackFunc(feature, ctx),
-			}
-		} else if opts.fallback != nil {
+		return handleFallback(opts, feature, ctx)
+	}
+
+	if f.Dependencies != nil && len(*f.Dependencies) > 0 {
+		parentEnabled := uc.isParentDependencySatisfied(f, *ctx)
```
kwasniew: isn't it more than parent enabled? the variable name may be misleading

client.go
```diff
@@ -254,31 +254,25 @@ func (uc *Client) isEnabled(feature string, options ...FeatureOption) api.Strate
...
 	if f == nil {
-		if opts.fallbackFunc != nil {
-			return api.StrategyResult{
-				Enabled: opts.fallbackFunc(feature, ctx),
-			}
-		} else if opts.fallback != nil {
+		return handleFallback(opts, feature, ctx)
+	}
+
+	if f.Dependencies != nil && len(*f.Dependencies) > 0 {
```
kwasniew, line 268: I'm curious why you decided to do this check outside the isParentDependencySatisfied. I made the opposite decision but didn't have a strong opinion here

kwasniew (approved):
Looks good!

client.go
FredrikOseberg: To me it made more sense readability wise. If I don't have dependencies or the length of dependencies is 0 then I don't need to do check if `isParentDependencySatisfied(f, *ctx)`. Moving it inside that function obfuscates the check that determines whether or not that function should actually be run. I don't have strong opinions about this either, but I found it easier to reason about this way because now the function is limited to doing one thing and one thing only.

client.go
FredrikOseberg: Agree

api/feature.go
FredrikOseberg: Dependency works fine

PR #143, opened 2023-10-09
fix: default to a noop listener to avoid hanging on channel messages

### LINKED ISSUE #139: infinitely blocked in metrics.count with gitlab
### Describe the bug

I am using Gitlab as unleash API, and I came across an error, where a request for an API-call against my microservice was timed out, since it never leaves the isEnabled func. After debugging I found out, that the metrics count in the defer is blocking infinitely.
https://github.com/Unleash/unleash-client-go/blob/v3/client.go#L243
Further it was really hard to figure this out as it is in the defer. I know, that I can disable the metrics, and that`s my fix for it, but imho this should not block. It should return an error instead or panic, but it was really hard to figure this out.

_No response_

### Expected behavior

No blocking

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

3.8.0

### Subscription type

None

### Hosting type

None

Go

daveleek:
## About the changes

Defaults to register a NoopListener if no listener has been added

Closes #139

## Discussion points

Should this also actually listen to the individual channels? I have commented that part out and seems to work without it but I'm unsure if this has longer term effects

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,15 @@ unleash.IsEnabled("someToggle", unleash.WithContext(ctx), unleash.WithResolver(r
 
 ## Development
 
+To override dependency on unleash-client-go github repository to a local development folder (for instance when building a local test-app for the SDK),  
+you can add the following to your apps `go.mod`:
+
+```mod
+    replace github.com/Unleash/unleash-client-go/v3 => ../unleash-client-go/
+```
+
+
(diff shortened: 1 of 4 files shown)


FredrikOseberg (approved):
LGTM

PR #145, opened 2023-10-26
fix: new variant hashing

sjaanus:
### What
Uses a new seed for ensuring a fair distribution for variants.

### Background
After a customer reported that variant distribution seemed skewed we performed some testing and found that since we use the same hash string for both gradual rollout and variant allocation we'd reduced the set of groups we could get to whatever percentage our gradual rollout was set.

### Example
Take a gradualRollout of 10%, this will select normalized hashes between 1 and 10, when we then again hash the same string that gave us between 1 and 10, but with modulo 1000 for variants, this will only give us 100 possible groups, instead of the expected 1000.

### Fix
Force the normalization to accept a seed, and make sure to use a new seed when normalizing the variant distribution hash.

### Worth noting
This will require release 4.0.0, since we are changing how hashing works.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
         name: Checkout client specifications
         with:
           repository: Unleash/client-specification
-          ref: refs/tags/v4.3.1
+          ref: refs/tags/v5.0.0
           path: testdata/client-specification
       - uses: actions/setup-go@v2
         name: Setup go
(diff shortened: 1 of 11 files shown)


chriswk:
Client Spec change is here: https://github.com/Unleash/client-specification/pull/68. It was only one failing client spec test for me

ivarconr (approved):

PR #146, opened 2023-10-27
fix: dependent feature toggle evaluation

sjaanus:
Upgrades client-spec to [4.5.2](https://github.com/Unleash/client-specification/releases/tag/v4.5.2) and fixes the logic


FredrikOseberg (approved):
LGTM

PR #149, opened 2023-11-20
chore: Revert breaking changes on v3

gastonfournier:
## About the changes
Revert changes on v3 that are major changes. They're already in v4 branch

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,7 +17,7 @@ import (
 const (
 	deprecatedSuffix = "/features"
 	clientName       = "unleash-client-go"
-	clientVersion    = "4.0.0"
+	clientVersion    = "4.0.0-beta.1"
 )
 
 var defaultStrategies = []strategy.Strategy{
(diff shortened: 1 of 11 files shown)


client.go
```diff
@@ -17,7 +17,7 @@ import (
 const (
 	deprecatedSuffix = "/features"
 	clientName       = "unleash-client-go"
-	clientVersion    = "4.0.0"
+	clientVersion    = "3.9.0"
```
kwasniew: should we do 3.9.1?

kwasniew (approved):
I'd only bump the patch version (especially if we leave the fix for dependent feature toggle evaluation)

client.go
gastonfournier: I was going to do this later as part of a release, but let me double check how is this done

gastonfournier:
```suggestion
	clientVersion    = "3.9.1"
```

PR #150, opened 2023-11-20
docs: improve onboarding

gastonfournier:
When getting started with go sdk it's easy to miss waiting for initialization. This change will make it easy for newcomers to use the SDK and more advanced users should be able to wait for the readiness in different ways if they don't want to block the main thread

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ func init() {
 		unleash.WithUrl("http://unleash.herokuapp.com/api/"),
 		unleash.WithCustomHeaders(http.Header{"Authorization": {"<API token>"}}),
 	)
+
+        // The line below will block until the default client is ready or returns immediately. You can wait for readiness in a different way but this is good for new joiners
+        unleash.WaitForReady()
 }
 ```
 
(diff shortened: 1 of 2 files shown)


kwasniew (commented):
This line makes me curious what are the other ways and why this is good for new joiners.

README.md
```diff
@@ -46,6 +46,9 @@ func init() {
 		unleash.WithUrl("http://unleash.herokuapp.com/api/"),
 		unleash.WithCustomHeaders(http.Header{"Authorization": {"<API token>"}}),
 	)
+
+        // The line below will block until the default client is ready or returns immediately. You can wait for readiness in a different way but this is good for new joiners
```
kwasniew: This line makes me curious what are the other ways and why this is good for new joiners.

gastonfournier: Most show different ways of initializing: sync vs async. Example: https://github.com/Unleash/unleash-client-java#example-configurations

Maybe we can use a similar format. I'm just waiting for confirmation that this was actually the problem

kwasniew (approved):

PR #154, opened 2023-11-22
test: separate specification tests from race conditions tests

kwasniew:
## About the changes

Separating specification tests (regular tests) and SDK internal tests (run with resource heavy race detection) using build tag I called `norace`.

Also found that some gradual rollout tests creating 200000 iterations which are a good candidate for exclusion from the race condition tests.


gastonfournier (approved):
LGTM! Because we're just tagging some specific tests most other tests will check for data races. :+1:

PR #156, opened 2023-11-29
feat: obey http status and backoff on 429, 50x

chriswk:
This follows the same pattern as Java and .NET.  each 429, 50x error will increment error count, and reset skips to error, then for each trigger reduce skips by 1 until you reach 0, then perform the action again. on success reduce error count by 1 and reset skips to error count.

On 401,403,404 will increment skips/errors to max at once and return an error saying you might've configured the client wrong

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"net/url"
 	"sync"
@@ -26,6 +27,9 @@ type repository struct {
 	isReady       bool
 	refreshTicker *time.Ticker
 	segments      map[int][]api.Constraint
+	errors        float64
+	maxSkips      float64
+	skips         float64
 }
 
 func newRepository(options repositoryOptions, channels repositoryChannels) *repository {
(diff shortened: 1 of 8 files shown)


repository.go
```diff
@@ -133,14 +162,21 @@ func (r *repository) fetch() error {
...
+func (r *repository) statusIsOK(resp *http.Response) error {
 	s := resp.StatusCode
 	if 200 <= s && s < 300 {
 		return nil
+	} else if s == 401 || s == 403 || s == 404 {
+		r.configurationError()
+		return fmt.Errorf("%s %s returned status code %d your SDK is most likely misconfigured, backing off to maximum (%f times our interval)", resp.Request.Method, resp.Request.URL, s, r.maxSkips)
+	} else if s == 429 || s >= 500 {
+		r.backoff()
+		return fmt.Errorf("%s %s returned status code %d, backing off (%f times our interval)", resp.Request.Method, resp.Request.URL, s, r.errors)
```
FredrikOseberg: Here we use the default HTTP codes, whereas in metrics we use the HTTP helpers. We should be consistent.

FredrikOseberg (approved):
LGTM. Small thing to fix, but feel free to merge once it's fixed.

repository.go
chriswk: Agreed, I'll update to using helpers here as well

chriswk: Fixed in 7c8994d

ivarconr:
💯

PR #157, opened 2023-11-30
chore: update to supported go versions

chriswk:
### Discussion
Having read https://go.dev/doc/devel/release I realised we haven't looked at our build matrix since 1.16 came out, and our go.mod is still pointing to 1.13 which was EoL 2021-02-16.

This PR bumps go.mod to 1.21 (the current oldest supported version) and updates the build matrix to 1.21, 1.22 and 1.23

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,5 @@
-github.com/Masterminds/semver/v3 v3.1.1 h1:hLg3sBzpNErnxhQtUy/mmLR2I9foDujNK030IGemrRc=
-github.com/Masterminds/semver/v3 v3.1.1/go.mod h1:VPu/7SZ7ePZ3QOrcuXROw5FAcLl4a0cBrbBpGY/8hQs=
+github.com/Masterminds/semver/v3 v3.3.1 h1:QtNSWtVZ3nBfk8mAOu/B6v7FMJ+NHTIgUPi7rj+4nv4=
+github.com/Masterminds/semver/v3 v3.3.1/go.mod h1:4V+yj/TJE1HU9XfppCwVMZq3I84lprf4nC11bSS5beM=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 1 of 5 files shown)


go.sum
```diff
@@ -1,16 +1,30 @@
...
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
-github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
-github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
-github.com/stretchr/testify v1.2.2 h1:bSDNvY7ZPG5RlJ8otE/7V6gMiyenm9RtJ7IUVIAoJ1w=
-github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
-github.com/twmb/murmur3 v1.1.5 h1:i9OLS9fkuLzBXjt6dptlAEyk58fJsSTXbRg3SgVyqgk=
-github.com/twmb/murmur3 v1.1.5/go.mod h1:Qq/R7NUyOfr65zD+6Q5IHKsJLwP7exErjN6lyyq3OSQ=
-gopkg.in/h2non/gock.v1 v1.0.10 h1:D4j796HhgidcxF0LnDyFXcoEbEZWoLEWf0kRh61p22w=
-gopkg.in/h2non/gock.v1 v1.0.10/go.mod h1:KHI4Z1sxDW6P4N3DfTWSEza07YpkQP7KJBfglRMEjKY=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
```
kwasniew: why do we have explosion of versions here?

chriswk: I believe that go 1.17 updated the format for go.mod/go.sum, and since we now explicitly state go 1.19 in our go.mod, the new format takes effect.

gastonfournier:
With this change we might be dropping support for older versions of go. Maybe something we want to do... I did test the current v4 against newer versions of go and it seems to work well: https://github.com/Unleash/unleash-client-go/pull/177

chriswk:
> With this change we might be dropping support for older versions of go. Maybe something we want to do... I did test the current v4 against newer versions of go and it seems to work well: #177

Go itself has an explicit "we only support until two new major versions have been released" policy

nunogois (approved):
LGTM

PR #158, opened 2023-12-06
chore: clean tests usage of endpoints

chriswk:
Previously, once we hit the feature refresh interval, we saw failed requests against https://localhost:4242, this update replaces all usages in these tests with the gocked url instead of a fake localhost address


daveleek (approved):
LGTM!

PR #164, opened 2023-12-08
feat: add FeatureEnabled to Variant

### LINKED ISSUE #159: Return variant and feature enabled from a single method

Currently there's no way with a single call to differentiate between a feature being disabled and no variants on the feature. Both of those cases return the default variant. Ideally there would be a call that returns `(variant, enabled)`.

### Background

We would like to make a single call and understand if a feature is enabled and what variant is applicable, if any. Instead we have to make 2 calls to understand the distinction. I'm surprised that this hasn't been brought up already and I'm curious how others are using this library in production with variants.

### Solution suggestions

There are 2 potential solutions I see:
1. Fixing the fallback options. Currently the descriptions of [WithVariantFallback](https://github.com/Unleash/unleash-client-go/blob/v4.0.1/config.go#L219) and [WithVariantFallbackFunc](https://github.com/Unleash/unleash-client-go/blob/v4.0.1/config.go#L227) are wrong. Those are actually used when a *feature* is not found. I filed this as https://github.com/Unleash/unleash-client-go/issues/160. A new option could be added that is used to fallback when a feature is not found and the existing fallbacks could be correctly used when there are no variants on the feature. This would be a breaking change though, but I'm not sure how many people realize the existing options are not doing what they say they do.
2. Add a new method that returns a variant and if it's enabled or not. This could also return a struct similar to the StrategyResult struct so it could be added to in the future if more things need to be returned.

Personally we're fine with either but the options being broken does seem like it should be fixed and might lend itself to just adding a new option.

jameshartig:
## About the changes

FeatureEnabled is similar to Enabled except in the case where the feature is enabled but there are no variants defined. This follows the client specification case [1].

[1] https://github.com/Unleash/client-specification/blob/c0169d7ace35db66cdf41a7b1b4e390a4a843c3b/specifications/08-variants.json#L448

Closes #159 

### Important files

api/variant.go

## Discussion points

It's a bit unclear what should happen with `api.DISABLED_VARIANT`. Personally I think `GetDefaultVariant` should be deprecated and instead users should be checking `Enabled` on the variant but I'm not sure what other SDKs are doing. I added a new `GetDefaultVariantFeatureEnabled` method to return the feature-enabled default variant but it's not obvious how anyone outside of the unleash client should be using the `GetDefaultVariant` method(s). There might need to be more clarification on the methods to discourage or guide usage.

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -89,6 +89,7 @@ func (vc VariantCollection) GetVariant(ctx *context.Context) *Variant {
 			variant = &v.Variant
 		}
 		variant.Enabled = true
+		variant.FeatureEnabled = true
 		return variant
 	}
 	return DISABLED_VARIANT
--- a/api/variant.go
+++ b/api/variant.go
@@ -3,8 +3,9 @@ package api
 import "github.com/Unleash/unleash-client-go/v4/context"
 
 var DISABLED_VARIANT = &Variant{
-	Name:    "disabled",
-	Enabled: false,
+	Name:           "disabled",
+	Enabled:        false,
+	FeatureEnabled: false,
 }
 
 type Payload struct {
@@ -26,8 +27,11 @@ type Variant struct {
 	Name string `json:"name"`
 	// Payload is the value of the variant payload
 	Payload Payload `json:"payload"`
-	// Enabled indicates whether the feature which is extend by this variant was enabled or not.
+	// Enabled indicates whether the variant is enabled. This is only false when
+	// it's a default variant.
 	Enabled bool `json:"enabled"`
+	// FeatureEnabled indicates whether the Feature for this variant is enabled.
+	FeatureEnabled bool `json:"featureEnabled"`
 }
 
 type VariantInternal struct {
@@ -81,7 +85,10 @@ func (o Override) matchValue(ctx *context.Context) bool {
 	return false
 }
 
-// Get default variant if no variant is found
+// Get default variant if feature is not found or if the feature is disabled.
+//
+// Rather than checking against this particular variant you should be checking
+// the returned variant's Enabled and FeatureEnabled properties.
 func GetDefaultVariant() *Variant {
 	return DISABLED_VARIANT
 }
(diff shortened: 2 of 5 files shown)


FredrikOseberg (approved):
@jameshartig 

Thank you! This looks good to me.

I agree that the disabled variant should be deprecated. Perhaps the GetDefaultVariant method could be enhanced to take in a parameter and dynamically create the disabled variant with the new FeatureEnabled property? Then we wouldn't have to maintain two methods and two statically defined variants. Would require some additional changes in `getVariantWithoutMetrics` though. 

I'm not actually sure why these methods are public. I can't see any good reason why they should be. I'll make a note of investigating if we can make them private in an upcoming major version.  

EDIT: Seems like we merged some adjustments to the tests. Can I ask you to resolve them when you have a free second? Thank you.

jameshartig:
> Perhaps the GetDefaultVariant method could be enhanced to take in a parameter and dynamically create the disabled variant with the new FeatureEnabled property?

I thought about that but it would be a breaking change. After thinking about it more, what do you think about removing the `GetDefaultVariantFeatureEnabled` I added to discourage use, especially if we plan to remove the existing method later?

I can move the new variant to not be defined in the `api` package so it's not public easier.

My thinking is just that it's always easier to expose something if its needed later than to do the opposite.

FredrikOseberg:
> > Perhaps the GetDefaultVariant method could be enhanced to take in a parameter and dynamically create the disabled variant with the new FeatureEnabled property?
> 
> I thought about that but it would be a breaking change. After thinking about it more, what do you think about removing the `GetDefaultVariantFeatureEnabled` I added to discourage use, especially if we plan to remove the existing method later?
> 
> I can move the new variant to not be defined in the `api` package so it's not public easier.
> 
> My thinking is just that it's always easier to expose something if its needed later than to do the opposite.

Absolutely. We'd have to bump to 5 in that case. Yeah, I think that's a good idea for now, let's keep the new stuff internal and we can always expose it later if it makes sense.

jameshartig:
@FredrikOseberg thanks for the quick replies! I updated the PR with those changes and added some clarification to the existing method.

FredrikOseberg:
Fantastic. Thank you! Merging this now.

PR #165, opened 2023-12-13
Revert "feat: add FeatureEnabled to Variant"

FredrikOseberg:
Reverts Unleash/unleash-client-go#164

Looks like the changes caused the spec tests to fail: https://github.com/Unleash/unleash-client-go/actions/runs/7178688374/job/19547365015

This is surprising because the workflow should run the spec tests when you setup a new PR, but for some reason this time they weren't triggered.

CC: @jameshartig 

Could you check if the spec tests pass locally with your changes? There should be instructions [here](https://github.com/Unleash/unleash-client-go?tab=readme-ov-file#adding-client-specifications). If not I can try to do that later.

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -89,7 +89,6 @@ func (vc VariantCollection) GetVariant(ctx *context.Context) *Variant {
 			variant = &v.Variant
 		}
 		variant.Enabled = true
-		variant.FeatureEnabled = true
 		return variant
 	}
 	return DISABLED_VARIANT
--- a/api/variant.go
+++ b/api/variant.go
@@ -3,9 +3,8 @@ package api
 import "github.com/Unleash/unleash-client-go/v4/context"
 
 var DISABLED_VARIANT = &Variant{
-	Name:           "disabled",
-	Enabled:        false,
-	FeatureEnabled: false,
+	Name:    "disabled",
+	Enabled: false,
 }
 
 type Payload struct {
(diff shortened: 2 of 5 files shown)


thomasheartman (approved):

jameshartig:
@FredrikOseberg Sorry about that! I didn't realize I needed to check out the client specs for those tests to run. I'll work on that now real quick.

jameshartig:
@FredrikOseberg I admittedly don't have a lot of experience with these tests but It looks like my code is actually working but interpreting the spec is broken.

```
=== RUN   TestClientSpecificationSuite/TestClientSpecification/17-dependent-features/Child_yields_its_variant_when_both_parent_and_child_are_enabled.
    spec_test.go:91:
                Error Trace:    spec_test.go:91
                Error:          Not equal:
                                expected: &api.Variant{Name:"child.variant", Payload:api.Payload{Type:"string", Value:"variantValue"}, Enabled:true, FeatureEnabled:false}
                                actual  : &api.Variant{Name:"child.variant", Payload:api.Payload{Type:"string", Value:"variantValue"}, Enabled:true, FeatureEnabled:true}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -7,3 +7,3 @@
                                  Enabled: (bool) true,
                                - FeatureEnabled: (bool) false
                                + FeatureEnabled: (bool) true
                                 })
                Test:           TestClientSpecificationSuite/TestClientSpecification/17-dependent-features/Child_yields_its_variant_when_both_parent_and_child_are_enabled.
```

But the spec for that says:
```
    {
      "description": "Child yields its variant when both parent and child are enabled.",
      "context": {},
      "toggleName": "parent.enabled.child.enabled",
      "expectedResult": {
        "name": "child.variant",
        "payload": {
          "type": "string",
          "value": "variantValue"
        },
        "enabled": true,
        "feature_enabled": true
      }
    }
```

I wonder if the issue is that `feature_enabled` is using underscores the `api.Variant`'s FeatureEnabled field is `featureEnabled`.

PR #166, opened 2023-12-13
feat: add FeatureEnabled to Variant and fix spec testing

### LINKED ISSUE #159: Return variant and feature enabled from a single method

jameshartig:
## About the changes

FeatureEnabled is similar to Enabled except in the case where the feature is enabled but there are no variants defined. This follows the client specification case [1].

[1] https://github.com/Unleash/client-specification/blob/c0169d7ace35db66cdf41a7b1b4e390a4a843c3b/specifications/08-variants.json#L448

Closes #159

## Discussion points

See https://github.com/Unleash/unleash-client-go/pull/165#issuecomment-1854245853 but I'm not sure how to handle the JSON differences between the spec and `*api.Variant`.

### DIFF
--- a/.github/workflows/build-prs.yml
+++ b/.github/workflows/build-prs.yml
@@ -1,9 +1,7 @@
-name: Build
+name: Build PRs
 
 on:
-  push:
-    branches:
-      - main
+  pull_request:
 
 jobs:
   build:
@@ -12,15 +10,15 @@ jobs:
       matrix:
         version: [1.13, 1.15, 1.16]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v2
         name: Checkout code
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v2
         name: Checkout client specifications
         with:
           repository: Unleash/client-specification
           ref: refs/tags/v5.1.0
           path: testdata/client-specification
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v2
         name: Setup go
         with:
           go-version: ${{ matrix.version }}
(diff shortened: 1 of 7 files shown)


jameshartig:
@FredrikOseberg took my best shot at fixing the spec tests and they seem to pass locally now.

FredrikOseberg:
@jameshartig Thanks. We don't have detailed documentation about our specification tests, but it's basically a set of tests that we use to ensure SDK compatibility across languages. Nice work on setting it up, finding the error and fixing it. 

I noticed a flaw in our current workflow that only built and ran the tests when a push was made by someone with write access to the repository, so I added a new workflow that builds PRs and changed the other workflow to run on main. All the checks pass now!

FredrikOseberg (approved):

PR #167, opened 2023-12-13
feat: return feature from isEnabled to avoid another lookup

### LINKED ISSUE #161: Avoid fetching feature again in GetVariant

Currently `GetVariant` calls `isEnabled` and then uses a custom resolver or `getToggle` to fetch the feature immediately afterwards.

### Background

We are starting to use the Unleash client in very performance-sensitive places and we notied this inefficiency in testing when we noticed our custom resolver being called twice.

### Solution suggestions

Ideally `StrategyResult` could contain the `Feature` so that `GetVariant` doesn't have to fetch it again. I do realize this is probably a small perf improvement but it seems simple.

jameshartig:
## About the changes

Previously GetVariant would do 2 storage.Get calls which might be costly depending on the implementation. It doesn't need to do the second one if we just return the feature from isEnabled.

Closes #161

### DIFF
--- a/client.go
+++ b/client.go
@@ -245,12 +245,13 @@ func (uc *Client) IsEnabled(feature string, options ...FeatureOption) (enabled b
 		uc.metrics.count(feature, enabled)
 	}()
 
-	return uc.isEnabled(feature, options...).Enabled
+	result, _ := uc.isEnabled(feature, options...)
+	return result.Enabled
 }
 
 // isEnabled abstracts away the details of checking if a toggle is turned on or off
 // without metrics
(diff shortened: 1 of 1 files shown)


FredrikOseberg:
> ## About the changes
> Previously GetVariant would do 2 storage.Get calls which might be costly depending on the implementation. It doesn't need to do the second one if we just return the feature from isEnabled.
> 
> Closes #161
> 
> ### Important files
> ## Discussion points

I'll take a look at this as soon as I can

FredrikOseberg (approved):
LGTM

PR #168, opened 2023-12-13
feat(api): add type to Feature

### LINKED ISSUE #162: Add Type to api.Feature

The API returns the `type` of the feature. It should be added as a struct field on `api.Feature`.

### Background

We will have some special metrics around `kill-switch` type features to specifically track when those are being enabled to kill features.

### Solution suggestions

Add `Type string` field to `api.Feature`.

jameshartig:
## About the changes

Adds the `Type` field to `api.Feature` which is already returned by the API but was not captured in the Go struct.

Closes #162

## Discussion points

This will be slightly difficult for us to get to in normal usage since `IsEnabled` and `GetVariant` do not expose the underlying `api.Feature`. We plan to add our own `map[string]*api.Feature` to our internal client so we can resolve the underlying feature after we call `GetVariant`.

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -26,6 +26,9 @@ type Feature struct {
 	// Name is the name of the feature toggle.
 	Name string `json:"name"`
 
+	// Type is the type of the feature toggle.
+	Type string `json:"type"`
+
 	// Description is a description of the feature toggle.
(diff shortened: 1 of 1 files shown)


FredrikOseberg (approved):
LGTM

PR #169, opened 2023-12-18
[wip] fix: variant fallback usage

### LINKED ISSUE #160: WithVariantFallback and WithVariantFallbackFunc are called when feature isn't found
### Describe the bug

[WithVariantFallback](https://github.com/Unleash/unleash-client-go/blob/v4.0.1/config.go#L219) and [WithVariantFallbackFunc](https://github.com/Unleash/unleash-client-go/blob/v4.0.1/config.go#L227) both say they are called when no variant is found but they're actually called when a *feature* isn't found. There's no way to provide a fallback when variants aren't found.

1. Call `GetVariant` on a non-existent feature with `WithVariantFallback`.
2. Observe fallback being used when the feature doesn't exist

### Expected behavior

I expected the options to match their descriptions and only be called when the variant isn't found. 

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

Pro

### Hosting type

Hosted by Unleash

Version 4.0.1

thomasheartman:
This PR fixes a bug in how we use fallback variants in this SDK.

The fallback should be used when there is no flag, when the flag doesn't have any variants, and if the flag has variants, but is disabled. However, prior to this, it was only used if the flag didn't exist.

It addresses the issues in and closes #160.

### Overriding `FeatureEnabled`

Comparing this to the Node SDK, I noticed one thing that we don't seem to do: setting the `FeatureEnabled` property based on the flag's enabled state. 

Because the fallback you pass in also has `FeatureEnabled` as a property, you can in theory override this. However, that means you can get into situations where the flag is actually enabled, but the variant says `FeatureEnabled` is `false` because that's what the fallback says. That seems incorrect to me, so I've added in some handling for that, but I'd like to get a second opinion on it.

### Code quality

This may be my first time writing Go, so it's very possible that I've missed some idioms or similar. Please feel free to correct as many things as you want.

### DIFF
--- a/client.go
+++ b/client.go
@@ -417,8 +417,13 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 
 	if !strategyResult.Enabled {
 		if opts.variantFallbackFunc != nil {
-			return opts.variantFallbackFunc(feature, ctx)
+			variant := opts.variantFallbackFunc(feature, ctx)
+			if variant != nil {
+				variant.FeatureEnabled = false
+			}
+			return variant
 		} else if opts.variantFallback != nil {
+			opts.variantFallback.FeatureEnabled = false
 			return opts.variantFallback
 		}
 		return defaultVariant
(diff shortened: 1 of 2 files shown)


thomasheartman:
closing in favor of #171

PR #171, opened 2023-12-18
fix: variant fallback usage

### LINKED ISSUE #160: WithVariantFallback and WithVariantFallbackFunc are called when feature isn't found

thomasheartman:
This PR fixes a bug in how we use fallback variants in this SDK.

The fallback should be used when there is no flag, when the flag doesn't have any variants, and if the flag has variants, but is disabled. However, prior to this, it was only used if the flag didn't exist.

It addresses the issues in and closes #160.

### Overriding `FeatureEnabled`

Comparing this to the Node SDK, I noticed one thing that we don't seem to do: setting the `FeatureEnabled` property based on the flag's enabled state.

Because the fallback you pass in also has `FeatureEnabled` as a property, you can in theory override this. However, that means you can get into situations where the flag is actually enabled, but the variant says `FeatureEnabled` is `false` because that's what the fallback says. That seems incorrect to me, so I've added in some handling for that, but I'd like to get a second opinion on it.

### Code quality

This may be my first time writing Go, so it's very possible that I've missed some idioms or similar. Please feel free to correct as many things as you want.

This includes the very verbose test setup which accounts for ~85% of this PR.

### DIFF
--- a/client.go
+++ b/client.go
@@ -418,13 +418,8 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 
 	getFallbackVariant := func(featureEnabled bool) *api.Variant {
 		if opts.variantFallbackFunc != nil {
-			variant := opts.variantFallbackFunc(feature, ctx)
-			if variant != nil {
-				variant.FeatureEnabled = featureEnabled
-			}
-			return variant
+			return opts.variantFallbackFunc(feature, ctx)
 		} else if opts.variantFallback != nil {
-			opts.variantFallback.FeatureEnabled = featureEnabled
 			return opts.variantFallback
 		}
 
--- a/client.go
+++ b/client.go
@@ -416,7 +416,7 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 		strategyResult, f = uc.isEnabled(feature, WithContext(*ctx))
 	}
 
-	getFallbackWithFeatureEnabled := func(featureEnabled bool) *api.Variant {
+	getFallbackVariant := func(featureEnabled bool) *api.Variant {
 		if opts.variantFallbackFunc != nil {
 			variant := opts.variantFallbackFunc(feature, ctx)
 			if variant != nil {
@@ -435,11 +435,11 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 	}
 
 	if !strategyResult.Enabled {
-		return getFallbackWithFeatureEnabled(false)
+		return getFallbackVariant(false)
 	}
 
 	if f == nil || !f.Enabled {
-		return getFallbackWithFeatureEnabled(false)
+		return getFallbackVariant(false)
 	}
 
 	if strategyResult.Variant != nil {
@@ -447,7 +447,7 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 	}
 
 	if len(f.Variants) == 0 {
-		return getFallbackWithFeatureEnabled(true)
+		return getFallbackVariant(true)
 	}
 
 	return api.VariantCollection{
--- a/client.go
+++ b/client.go
@@ -416,21 +416,30 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 		strategyResult, f = uc.isEnabled(feature, WithContext(*ctx))
 	}
 
-	if !strategyResult.Enabled {
-		return defaultVariant
-	}
-
-	if f == nil {
+	getFallbackWithFeatureEnabled := func(featureEnabled bool) *api.Variant {
 		if opts.variantFallbackFunc != nil {
-			return opts.variantFallbackFunc(feature, ctx)
+			variant := opts.variantFallbackFunc(feature, ctx)
+			if variant != nil {
+				variant.FeatureEnabled = featureEnabled
+			}
+			return variant
 		} else if opts.variantFallback != nil {
+			opts.variantFallback.FeatureEnabled = featureEnabled
 			return opts.variantFallback
+		}
+
+		if featureEnabled {
+			return disabledVariantFeatureEnabled
 		}
 		return defaultVariant
 	}
 
-	if !f.Enabled {
-		return defaultVariant
+	if !strategyResult.Enabled {
+		return getFallbackWithFeatureEnabled(false)
+	}
+
+	if f == nil || !f.Enabled {
+		return getFallbackWithFeatureEnabled(false)
 	}
 
 	if strategyResult.Variant != nil {
@@ -438,7 +447,7 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
 	}
 
 	if len(f.Variants) == 0 {
-		return disabledVariantFeatureEnabled
+		return getFallbackWithFeatureEnabled(true)
 	}
 
 	return api.VariantCollection{
--- a/client_test.go
+++ b/client_test.go
@@ -1180,17 +1180,11 @@ func TestGetVariantWithFallbackVariantWhenFeatureDisabled(t *testing.T) {
 			Segments: []api.Segment{},
 		})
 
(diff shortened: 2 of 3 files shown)


client.go
```diff
@@ -416,29 +416,38 @@ func (uc *Client) getVariantWithoutMetrics(feature string, options ...VariantOpt
...
 		}
+
+		if featureEnabled {
+			return disabledVariantFeatureEnabled
+		}
 		return defaultVariant
 	}
 
-	if !f.Enabled {
-		return defaultVariant
```
thomasheartman, line 433: I don't know when we ever hit this? I couldn't find a way to trigger it. One of the tests uses a default strategy (which is always enabled), but with `f.Enabled` set to false. This seems to still hit the previous `strategyResult.enabled`, though, so I suspect this isn't used anymore, but I'll leave it in combined with missing flags regardless.

client_test.go
```diff
@@ -1144,3 +1144,216 @@ func TestClient_VariantFromEnabledFeatureWithNoVariants(t *testing.T) {
...
+			Features: features,
+			Segments: []api.Segment{},
+		})
+
+	mockListener := &MockedListener{}
+	mockListener.On("OnReady").Return()
+	mockListener.On("OnRegistered", mock.AnythingOfType("ClientData"))
+	mockListener.On("OnCount", feature, false).Return()
+	mockListener.On("OnError").Return()
+
```
chriswk: unless you explicitly want to test something with the Listener, you can use the NoopListener instead of a mocked listener here.

chriswk (approved):
If you were afraid of non idiomatic go, I might not be the best reviewer, but this certainly looks good to me. I understand what you're doing and I like the added tests.

:+1:

config.go
```diff
@@ -214,16 +214,20 @@ type variantOption struct {
 // VariantOption provides options for querying if a variant is found or not.
 type VariantOption func(*variantOption)
 
-// WithVariantFallback specifies what the value should be if the variant is not found on the
-// unleash service.
+// WithVariantFallback specifies what the value should be if the
```
jameshartig, line 217: I think it's useful to say that the `FeatureEnabled` field will be overwritten in this comment and the next one.

client_test.go
```diff
@@ -1144,3 +1144,216 @@ func TestClient_VariantFromEnabledFeatureWithNoVariants(t *testing.T) {
...
+		WithInstanceId(mockInstanceId),
+		WithListener(mockListener),
+	)
+
+	assert.NoError(err)
+	client.WaitForReady()
+
+	fallbackVariant := api.Variant{
+		Name: "fallback-variant",
+                FeatureEnabled: true,
```
jameshartig: Spacing is a bit weird here

client_test.go
```diff
@@ -1144,3 +1144,216 @@ func TestClient_VariantFromEnabledFeatureWithNoVariants(t *testing.T) {
...
+
+	assert.Equal(fallbackVariant, *variant)
+
+	fallbackFunc := func(feature string, ctx *context.Context) *api.Variant {
+		return &fallbackVariant
+	}
+
+	variantWithFallbackFunc := client.GetVariant(feature, WithVariantFallbackFunc(fallbackFunc))
+
+	assert.Equal(fallbackVariant, *variantWithFallbackFunc)
```
jameshartig, line 1211: While I get that the `FeatureEnabled` field is being changed in the previous call to `GetVariant` that is a bit confusing here looking at the test. I think it might be better if you did:

```go
	fallbackVariant = api.Variant{
		Name: "fallback-variant",
    FeatureEnabled: true,
	}
	fallbackFunc := func(feature string, ctx *context.Context) *api.Variant {
		return &fallbackVariant
	}
```

then later did the same
```go
	assert.False(variantWithFallbackFunc.Enabled)

	assert.False(variantWithFallbackFunc.FeatureEnabled)
```

client_test.go
```diff
@@ -1144,3 +1144,216 @@ func TestClient_VariantFromEnabledFeatureWithNoVariants(t *testing.T) {
...
```
jameshartig, line 1283: Same thing here just to make it clear what the test is testing in the second case

client_test.go
```diff
@@ -1144,3 +1144,216 @@ func TestClient_VariantFromEnabledFeatureWithNoVariants(t *testing.T) {
...
```
jameshartig, line 1337: Same thing in this test too

jameshartig:
I left a few comments. I do think the `FeatureEnabled` overriding makes sense but since it is a pointer it's a bit risky. There definitely needs to be something in those method docs that makes it very clear that the pointer will be modified.

jameshartig:
There could actually be a race here where a caller passes in the same pointer to multiple calls to `GetVariant` and that ends up triggering a race when setting that field. Because of this possible race I'm actually leaning the other way now and maybe think we should just return it as-is (without touching the variant) at least until we have a solution for the race. @thomasheartman @chriswk thoughts?

thomasheartman:
@jameshartig Appreciate the review! I'll address your comments. As for touching / not touching the `FeatureEnabled` field. Yeah, that's a good point, which is why I was a little wary of it. Because we're dealing with pointers here (as far as I can tell), it makes it trickier to return a struct with just that field overridden. I'm happy to leave out changing the `FeatureEnabled` field (but I think that should also be clarified in the function comments and the docs).

client_test.go
thomasheartman: Ah, that's nice. Thanks!

thomasheartman: Addressed in d4a5c323588dedcddca9936707e5d0f698ee7be4

client_test.go
thomasheartman: That's probably a tabs/spaces thing. Weird that my editor changed to using whatever the other is. Anyway, that line is gone now, so that's not an issue anymore.

client_test.go
thomasheartman: Removed changing the feature enabled in 251d167c134e70a2f58ce5420b88064bc7daaae3 as per your suggestion. With that change, is there still anything here you'd like to address?

thomasheartman:
I changed the `FeatureEnabled` handling in 251d167c134e70a2f58ce5420b88064bc7daaae3 and added some extra tests for that explicitly in bc87bd3e8f04a05828a84b6e5ec1367c6e814c3d. Does that seem sensible to you @jameshartig? Oh, and I'd be very interested to hear your thoughts on that handling as well, @chriswk ☺️

client_test.go
jameshartig: Nope, lgtm

jameshartig (approved):

jameshartig:
> I changed the `FeatureEnabled` handling in [251d167](https://github.com/Unleash/unleash-client-go/commit/251d167c134e70a2f58ce5420b88064bc7daaae3) and added some extra tests for that explicitly in [bc87bd3](https://github.com/Unleash/unleash-client-go/commit/bc87bd3e8f04a05828a84b6e5ec1367c6e814c3d). Does that seem sensible to you @jameshartig? Oh, and I'd be very interested to hear your thoughts on that handling as well, @chriswk ☺️

Yep! It looks good now. I do agree that it's unfortunate the field isn't being automatically set but that might be something we can solve down the road in a v5.

PR #172, opened 2023-12-21
task: Bumped clientVersion

chriswk:
As the title says, bumps clientVersion so we can tag 4.1.0


chriswk:
how is our coverage reduced by me bumping a clientVersion?

daveleek (approved):
LG!

PR #174, opened 2024-07-10
feat: automatically generated instanceId

nunogois:
https://linear.app/unleash/issue/2-2395/make-instance-id-autogenned-go

This makes `instanceId` an automatically generated property, instead of being set by the user.

Similar to: 
 - https://github.com/Unleash/unleash-client-ruby/pull/179
 - https://github.com/Unleash/unleash-client-node/pull/640
 - https://github.com/Unleash/unleash-client-dotnet/pull/226


sighphyre (approved):
LGTM

PR #175, opened 2024-07-26
feat: New Metrics metadata

chriswk:
is included in client registration and client metrics

### DIFF
--- a/config.go
+++ b/config.go
@@ -108,11 +108,6 @@ func WithStrategies(strategies ...strategy.Strategy) ConfigOption {
 		o.strategies = strategies
 	}
 }
-func WithStarted(startedAt time.Time) ConfigOption {
-	return func(o *configOption) {
-		o.started = &startedAt
-	}
-}
 
 // WithStorage specifies which storage implementation the repository should use for storing feature
 // toggles.
--- a/config.go
+++ b/config.go
@@ -25,7 +25,6 @@ type configOption struct {
 	storage         Storage
 	httpClient      *http.Client
 	customHeaders   http.Header
-	started         *time.Time
 }
 
 // ConfigOption represents a option for configuring the client.
--- a/config.go
+++ b/config.go
@@ -25,6 +25,7 @@ type configOption struct {
 	storage         Storage
 	httpClient      *http.Client
 	customHeaders   http.Header
+	started         *time.Time
 }
 
 // ConfigOption represents a option for configuring the client.
@@ -105,6 +106,11 @@ func WithBackupPath(backupPath string) ConfigOption {
 func WithStrategies(strategies ...strategy.Strategy) ConfigOption {
 	return func(o *configOption) {
 		o.strategies = strategies
+	}
+}
+func WithStarted(startedAt time.Time) ConfigOption {
+	return func(o *configOption) {
+		o.started = &startedAt
 	}
 }
 
@@ -265,4 +271,5 @@ type metricsOptions struct {
 	disableMetrics  bool
 	httpClient      *http.Client
 	customHeaders   http.Header
+	started         *time.Time
 }
--- a/metrics.go
+++ b/metrics.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"net/http"
 	"net/url"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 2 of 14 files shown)


sighphyre (changes requested):
The code looks cool but what happened to `specVersion`?

config.go
```diff
@@ -107,6 +108,11 @@ func WithStrategies(strategies ...strategy.Strategy) ConfigOption {
 		o.strategies = strategies
 	}
 }
+func WithStarted(startedAt time.Time) ConfigOption {
+	return func(o *configOption) {
+		o.started = &startedAt
+	}
+}
```
nunogois: Should `started` be configurable?

chriswk: Mostly for tests here

chriswk: I would've used #[cfg(test)] here, don't know what the same is in go.

chriswk: Good point, I've moved it into the test file where I used it instead.

chriswk:
> The code looks cool but what happened to `specVersion`?

Don't you come here with your facts :1st_place_medal: . Added specVersion...

config.go
nunogois: I still don't like that we're polluting our config options with something that shouldn't be user-configurable and I'm still not super convinced of its value. Maybe we can relax our tests instead?

nunogois: Something like this seems to work, but maybe there's a better way:

```go
func clientDataMatcher() func(req *http.Request, ereq *gock.Request) (bool, error) {
	defaultStrategies := []string{
		"default",
		"applicationHostname",
		"gradualRolloutRandom",
		"gradualRolloutSessionId",
		"gradualRolloutUserId",
		"remoteAddress",
		"userWithId",
		"flexibleRollout",
	}
	return func(req *http.Request, ereq *gock.Request) (bool, error) {
		var data ClientData
		err := json.NewDecoder(req.Body).Decode(&data)
		if err != nil {
			return false, err
		}

		if data.Started.IsZero() {
			return false, nil
		}

		expectedData := ClientData{
			AppName:          mockAppName,
			InstanceID:       mockInstanceId,
			SDKVersion:       fmt.Sprintf("%s:%s", clientName, clientVersion),
			Strategies:       defaultStrategies,
			Interval:         0,
			PlatformVersion:  runtime.Version(),
			PlatformName:     "go",
			YggdrasilVersion: nil,
			SpecVersion:      specVersion,
		}

		return data.AppName == expectedData.AppName &&
			data.InstanceID == expectedData.InstanceID &&
			data.SDKVersion == expectedData.SDKVersion &&
			compareStringSlices(data.Strategies, expectedData.Strategies) &&
			data.Interval == expectedData.Interval &&
			data.PlatformVersion == expectedData.PlatformVersion &&
			data.PlatformName == expectedData.PlatformName &&
			data.YggdrasilVersion == expectedData.YggdrasilVersion &&
			data.SpecVersion == expectedData.SpecVersion, nil
	}
}

func compareStringSlices(a, b []string) bool {
	if len(a) != len(b) {
		return false
	}
	for i := range a {
		if a[i] != b[i] {
			return false
		}
	}
	return true
}

func TestMetrics_ClientDataIncludesNewMetadata(t *testing.T) {
	assert := assert.New(t)
	defer gock.OffAll()

	gock.New(mockerServer).
		Post("/client/register").
		AddMatcher(clientDataMatcher()).
		Reply(200)

	client, err := NewClient(
		WithUrl(mockerServer),
		WithMetricsInterval(50*time.Millisecond),
		WithAppName(mockAppName),
		WithInstanceId(mockInstanceId),
	)

	assert.Nil(err, "Client should open without a problem")

	time.Sleep(100 * time.Millisecond)

	err = client.Close()

	assert.Nil(err, "Client should close without errors")

	st.Expect(t, gock.IsDone(), true)
}
```

sighphyre (approved):
Noice! LGTM

config.go
chriswk: Thanks. I added your suggestion

chriswk: And removed started from the options

metrics.go
```diff
@@ -73,10 +98,14 @@ type metrics struct {
 }
 
 func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
+	started := time.Now()
+	if options.started != nil {
+		started = *options.started
+	}
```
nunogois: I believe this can be reverted now, and we can set `started: time.Now()` like before.

metrics_test.go
```diff
@@ -426,3 +429,166 @@ func TestMetrics_ErrorCountShouldDecreaseIfSuccessful(t *testing.T) {
...
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
```
nunogois, line 489: These are helper functions, not tests. Maybe we should move them to the top of the file or somewhere else that's not here?

nunogois (approved):
LGTM - Thank you for considering my comments 😄

metrics_test.go
chriswk: Why not here, they're used immediately after?

nunogois: Simply readability / convention / preference. I don't have strong feelings either way, and I understand the logic of keeping them near to where they're used, so I'm happy with keeping this.

PR #180, opened 2024-09-19
chore: update comment

thomasheartman:
Relates to a related fix done in some other sdks, such as the node one at
Unleash/unleash-client-node#417

In this case, the comment was misleading, so we can edit it.


sighphyre (approved):
LGTM

PR #186, opened 2025-01-27
feat: client identification headers

Tymek:
Identification headers consistent with other Unleash SDKs.

> This PR adds standardized client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
> - `unleash-appname`: the name of the application that is using the client. `UNLEASH-APPNAME` will be deleted in another PR (expand/contract pattern)
> - `unleash-connection-id`: an internal unique identifier for the current instance of the client generated by the built-in crypto lib
> - `unleash-sdk`: sdk information in the format `unleash-client-<language>:<version>`
> 
> All the headers are intended for the Unleash team. Changes should be implemented in a way that does not change SDK behavior in a significant way.
> The main use cases we have are:
> * capacity planning by knowing the number of unique connections made to the backend API
> * debugging misconfigured SDKs sending more traffic than expected

Fixes [issue/1-3271](https://linear.app/unleash/issue/1-3271/uniqueness-headers-in-go-sdk)

### DIFF
--- a/client.go
+++ b/client.go
@@ -169,9 +169,9 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 	if uc.options.customHeaders != nil {
 		headers = uc.options.customHeaders
 	}
-	headers.Set("x-unleash-appname", uc.options.appName)
-	headers.Set("x-unleash-sdk", fmt.Sprintf("%s:%s", clientName, clientVersion))
-	headers.Set("x-unleash-connection-id", getConnectionId())
+	headers.Set("unleash-appname", uc.options.appName)
+	headers.Set("unleash-sdk", fmt.Sprintf("%s:%s", clientName, clientVersion))
+	headers.Set("unleash-connection-id", getConnectionId())
 
 	uc.repository = newRepository(
 		repositoryOptions{
--- a/client.go
+++ b/client.go
@@ -3,6 +3,7 @@ package unleash
 import (
 	"fmt"
 
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
@@ -164,7 +165,10 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 		uc.options.instanceId = generateInstanceId()
 	}
 
-	headers := uc.options.customHeaders
+	headers := make(http.Header)
+	if uc.options.customHeaders != nil {
+		headers = uc.options.customHeaders
+	}
 	headers.Set("x-unleash-appname", uc.options.appName)
 	headers.Set("x-unleash-sdk", fmt.Sprintf("%s:%s", clientName, clientVersion))
 	headers.Set("x-unleash-connection-id", getConnectionId())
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -164,6 +165,14 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 		uc.options.instanceId = generateInstanceId()
 	}
 
+	headers := make(http.Header)
+	if uc.options.customHeaders != nil {
+		headers = uc.options.customHeaders
+	}
+	headers.Set("x-unleash-appname", uc.options.appName)
```
kwasniew: we can commit it without x instead of fixing it later

utils_test.go
```diff
@@ -116,3 +116,35 @@ func TestContains(t *testing.T) {
 		}
 	})
 }
+
+func TestGetConnectionId(t *testing.T) {
+	for i := 0; i < 100; i++ {
```
kwasniew, line 121: why do we run it 100 times? is there something specific about this SDK compared to the others where we test in only once?

utils.go
```diff
@@ -30,6 +31,24 @@ func generateInstanceId() string {
 	return prefix
 }
 
+func getConnectionId() string {
```
kwasniew, line 36: can we link to where we got this from. It's not a typical code someone writes every day

utils_test.go
Tymek: In order to avoid accidentally passing some tests, because implementation of "make everything random" could accidentally create a version and variant with correct values.

utils.go
Tymek: It is a new implementation. It's similar to
https://github.com/google/uuid/blob/2d3c2a9cc518326daf99a383f07c4d3c44317e4d/version4.go#L47-L56
and
https://github.com/hprose/hprose-go/blob/83de97da5004027694d321ca38c80fca3fac98c2/uuid.go#L91-L98
(because it does similar things).

client.go
Tymek: done

utils.go
kwasniew: I'd keep the link in the code as a comment

kwasniew (approved):

utils.go
Tymek, line 36:
```suggestion
// https://github.com/google/uuid/blob/2d3c2a9cc518326daf99a383f07c4d3c44317e4d/version4.go#L47-L56
// https://github.com/hprose/hprose-go/blob/83de97da5004027694d321ca38c80fca3fac98c2/uuid.go#L91-L98
func getConnectionId() string {
```

Tymek, line 36:
```suggestion
// https://github.com/google/uuid/blob/2d3c2a9cc518326daf99a383f07c4d3c44317e4d/version4.go#L47-L56
// https://github.com/hprose/hprose-go/blob/83de97da5004027694d321ca38c80fca3fac98c2/uuid.go#L91-L98
func getConnectionId() string {
```

PR #189, opened 2025-03-04
feat: add interval to header and connection Id to body

sjaanus:
1. Metrics - add connection id to payload and interval to header
2. Features - add interval to header
3. Register - add connection id to payload.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1491,7 +1491,7 @@ func TestConnectionAndIntervalHeadersAndBody(t *testing.T) {
 
 	client.IsEnabled("foo")
 
-	time.Sleep(320 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	err = client.Close()
 
 	assert.True(gock.IsDone(), "there should be no more mocks")
--- a/client_test.go
+++ b/client_test.go
@@ -1450,3 +1450,49 @@ func TestSendIdentificationHeaders(t *testing.T) {
 
 	assert.True(gock.IsDone(), "there should be no more mocks")
 }
+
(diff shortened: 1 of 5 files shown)


metrics.go
```diff
@@ -258,6 +265,7 @@ func (m *metrics) doPost(url *url.URL, payload interface{}) (*http.Response, err
 	req.Header.Add("UNLEASH-APPNAME", m.options.appName)
 	req.Header.Add("UNLEASH-INSTANCEID", m.options.instanceId)
 	req.Header.Add("User-Agent", m.options.appName)
+	req.Header.Add("Unleash-Interval", fmt.Sprintf("%d", m.options.metricsInterval.Milliseconds()))
```
kwasniew, line 268: it's weird that UNLEASH-INSTANCEID is all uppercase and this one is Unleash-Interval

client_test.go
```diff
@@ -1450,3 +1450,49 @@ func TestSendIdentificationHeaders(t *testing.T) {
...
+		WithInstanceId(mockInstanceId),
+		WithListener(&NoopListener{}),
+		WithDisableMetrics(false),
+	)
+
+	assert.NoError(err)
+
+	client.IsEnabled("foo")
+
+	time.Sleep(320 * time.Millisecond)
```
kwasniew: is it normal way to write tests in go? can't we wait until some event?

kwasniew (approved):

metrics.go
sjaanus: It is just industry standard to have headers as titled case.

PR #193, opened 2025-04-03
Add support for OnUpdate hook

### LINKED ISSUE #192: OnChange Listener

I am attempting to support switching the log level for my application via an unleash flag.  However unless there is a different approach to dealing with update hooks besides implementing my own timer, I am overlooking it.

I would like to use a callback such as OnChange or even OnUpdate to detect when flags are modified instead of testing for the flag in every log call regardless of the metrics.

Do you have any ideas on how to approach this?

### Background

_No response_

### Solution suggestions

_No response_

skiz:
## About the changes

Add a hook for identifying when the repository is updated so that further actions can be taken on the application side. Currently there is no way to trigger this without setting up another timer and re-evaluating a long lived flag to see if it has changed.

Useful for long lived state changes in applications such as being able to toggle log levels for short time periods without having to wrap each log with a flag check.

Closes #192 

## Discussion points

This may break the v4 API as it introduces a new listener method.

### DIFF
--- a/client.go
+++ b/client.go
@@ -53,6 +53,8 @@ type Client struct {
 	repositoryListener RepositoryListener
 	ready              chan bool
 	onReady            chan struct{}
+	update             chan bool
+	onUpdate           chan struct{}
 	close              chan struct{}
 	closed             chan struct{}
 	count              chan metric
@@ -76,7 +78,8 @@ func (ec errorChannels) err(err error) {
 
 type repositoryChannels struct {
 	errorChannels
-	ready chan bool
+	ready  chan bool
+	update chan bool
 }
 
 type metricsChannels struct {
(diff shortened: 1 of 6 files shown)


nunogois (changes requested):
This looks solid overall—nice work! One thing I wanted to flag: I believe Copilot’s suggestion about the `close(uc.onUpdate)` line is spot-on. Since onUpdate is intended to be used for future updates, we shouldn’t close the channel here. Closing it could lead to panics or missed updates down the line.

It would also be great to include some tests for this if possible. For example, a test involving multiple updates could help catch the issue mentioned above and guard against regressions in the future.

If you’re up for addressing those two points, I’d be happy to merge this in. 

Thanks again for your contribution!

PR #194, opened 2025-04-08
feat: impression data

### LINKED ISSUE #182: Impression Data

Some of our SDKs support [impression data](https://docs.getunleash.io/reference/impression-data)

The SDKs that have this support emits events when calling an evaluator (isEnabled | getVariant) and provides the surrounding context field values of this call and the result. This functionality was first introduced in the NodeJS SDK [here](https://github.com/Unleash/unleash-client-node/pull/336). 

We would like to also support this functionality in the golang SDK.

nunogois:
https://linear.app/unleash/issue/2-3494/go-sdk-impression-event-support

Implements impression data in our Go SDK, similar to what we already have for other SDKs.

Closes https://github.com/Unleash/unleash-client-go/issues/182


sighphyre (approved):
Seems sane to me, with my very limited Golang knowledge, nice one

Could use some docs but probably in another PR anyway

api/variant_test.go
```diff
@@ -331,3 +331,13 @@ func (suite *VariantTestSuite) TestGetVariant_OverrideOnEnvironment() {
...
+
+func TestGetDefaultVariant(t *testing.T) {
+	defaultVariant := GetDefaultVariant()
+	if defaultVariant.Name != "disabled" {
+		t.Errorf("Expected default variant name to be 'disabled', got '%s'", defaultVariant.Name)
+	}
+	if defaultVariant.Enabled {
+		t.Error("Expected default variant to be disabled")
+	}
+}
```
nunogois, line 343: Scouting / increasing our code coverage to make our GH workflow checks happy.

PR #196, opened 2025-05-28
Add support for OnUpdate hook

nathan-fiscaletti:
This is what I believe were the requested changes in https://github.com/Unleash/unleash-client-go/pull/193. It also addresses some (what I believe to be) issues in the original PR, such as the unnecessary declaration and assignment of the `onUpdate` channel. The only required channel is the `update` channel that was actually implemented.

The original issue addressed by this PR is #192.

I'm posting the original PR description below.

---

> ### Describe the feature request
> 
> I am attempting to support switching the log level for my application via an unleash flag. However unless there is a different approach to dealing with update hooks besides implementing my own timer, I am overlooking it.
> 
> I would like to use a callback such as OnChange or even OnUpdate to detect when flags are modified instead of testing for the flag in every log call regardless of the metrics.
> 
> Do you have any ideas on how to approach this?
> ### Background
> 
> _No response_
> ### Solution suggestions
> 
> _No response_

### DIFF
--- a/client.go
+++ b/client.go
@@ -53,6 +53,7 @@ type Client struct {
 	repositoryListener RepositoryListener
 	ready              chan bool
 	onReady            chan struct{}
+	update             chan bool
 	close              chan struct{}
 	closed             chan struct{}
 	count              chan metric
@@ -76,7 +77,8 @@ func (ec errorChannels) err(err error) {
 
 type repositoryChannels struct {
 	errorChannels
-	ready chan bool
+	ready  chan bool
+	update chan bool
 }
 
 type metricsChannels struct {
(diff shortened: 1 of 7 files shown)


chriswk (approved):
Hi Nathan.

One thing worth addressing, this will send an update signal for every fetch, even if the response was a 304 (no change). Just making sure that is what you wanted.

I think some of our other SDKs has two separate events, one that always triggers (so what you've written) and one that only triggers if the feature check actually returned 200.

I'm going to go ahead and approve what you have done here and again; thank you for going to the trouble :)

nathan-fiscaletti:
@chriswk I pushed one more update that addresses that issue with a new error value.

chriswk (approved):
Nice. Even better I think. Great work. One last ask, I would like a test confirming the new behaviour. 

I see our CI build flagged this as red due to the decreased test coverage.

nathan-fiscaletti:
@chriswk I just updated `repository_test.go` with a test that should cover this new logic. Let me know if it looks sane.

nathan-fiscaletti:
@chriswk any update on this?

nathan-fiscaletti:
@chriswk Pinging again, would like to get this merged in.

nathan-fiscaletti:
@chriswk Haven't heard anything in a week, hoping we can get this in

chriswk:
Yeah. This is all me, sorry. Merged now

PR #204, opened 2025-09-06
fix: Client.Ready() blocks forever when called after client is already ready

### LINKED ISSUE #203: Client.Ready() blocks forever when called after client is already ready
### Describe the bug

Summary
  - Client.Ready() blocks forever when called after client is already ready
  - Affected: v5.0.3 and earlier
  - Scope: Client.Ready() only; WaitForReady() and listener callbacks work as expected
  - Symptom: Calling Client.Ready() after the client has already become ready returns a channel that never receives, causing goroutines to hang indefinitely
  - Expected: Client.Ready() should resolve immediately (or otherwise deliver a value) even if readiness happened earlier

Minimal program that shows the hang when calling Ready() after the client is already ready.
```
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"
	"time"

	unleash "github.com/Unleash/unleash-go-sdk/v5"
)

func main() {
	// Mock Unleash server that immediately returns an empty features response.
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.URL.Path == "/client/features" {
			w.Header().Set("Content-Type", "application/json")
			w.WriteHeader(http.StatusOK)
			_, _ = w.Write([]byte(`{}`))
			return
		}
		http.NotFound(w, r)
	}))
	defer srv.Close()

	client, err := unleash.NewClient(
		unleash.WithUrl(srv.URL),
		unleash.WithAppName("app"),
		unleash.WithInstanceId("test"),
		unleash.WithDisableMetrics(true),
	)
	if err != nil {
		panic(err)
	}
	defer client.Close()

	// Ensure the client is ready.
	client.WaitForReady()
	fmt.Println("client is ready")

	// Reproduce: call Ready() AFTER readiness occurred.
	select {
	case <-client.Ready():
		fmt.Println("this should print, but on v5.0.3 and earlier it can hang")
	case <-time.After(100 * time.Millisecond):
		panic("timeout waiting for unleash client to be ready after 100ms")
	}
}
``` 

### Expected behavior

From reproduce code, the program will panic this message `panic: timeout waiting for unleash client to be ready after 100ms`
```
  	// Reproduce: call Ready() AFTER readiness occurred.
  	select {
  	case <-client.Ready():
  		fmt.Println("This should print, but on v5.0.3 it can hang")
  	case <-time.After(100 * time.Millisecond):
  		panic("timeout waiting for unleash client to be ready after 100ms")
  	}
```

### Logs, error output, etc.

```bash

```

### Screenshots

_No response_

### Additional context

I will propose the the pull request for workaround fix this issue.

### Unleash version

v7.2.0

### Subscription type

Open source

### Hosting type

Self-hosted

unleash-client-go v5.0.3 and eariler

kitimark:
## About the changes

Client.Ready() no longer exposes the internal one-shot `uc.ready` channel. It now returns a fresh channel per call that resolves by waiting on `uc.onReady` (a close-only broadcast). This ensures `Ready()` delivers even when called after the client has already become ready.

  - Affected versions: v5.0.3 and earlier
  - Unaffected APIs: `WaitForReady()` and listener callbacks

Closes #203

### Important files

- `client.go` — change to `Ready()` to use `uc.onReady` and per-call channel.
- `client_test.go` — tests for initial and post-ready calls with timeouts.

## Discussion points

- Currently repository (and in streaming, the delta processor) can signal readiness into `uc.ready` more than once. Closing uc.onReady once makes duplicate sources harmless, but confirm this is the intended single “ready” moment across modes.
- Returning a fresh channel per call changes channel identity. Unlikely anyone relied on the old identity, but it follows the functionality from docs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -548,11 +548,17 @@ func (uc *Client) Warnings() <-chan error {
 	return uc.warnings
 }
 
-// Ready returns the ready channel for the client. A value will be available on
-// the channel when the feature toggles have been loaded from the Unleash
-// server.
+// Ready returns a channel that will receive `true` when the client has loaded
+// feature toggles from the Unleash server. It returns a fresh channel per call
+// and will deliver immediately if the client is already ready.
 func (uc *Client) Ready() <-chan bool {
-	return uc.ready
+	ch := make(chan bool, 1)
+	go func() {
+		<-uc.onReady
+		ch <- true
+		close(ch)
+	}()
+	return ch
 }
 
 // Count returns the count channel which gives an update when a toggle has been queried.
--- a/client_test.go
+++ b/client_test.go
@@ -1496,3 +1496,86 @@ func TestConnectionAndIntervalHeadersAndBody(t *testing.T) {
 
 	assert.True(gock.IsDone(), "there should be no more mocks")
 }
+
+func TestReady(t *testing.T) {
+	// waitReadyChannelWithTimeout is a safety net for waiting on the ready channel to send the ready state.
+	// In the worst case, if the channel never sends anything after this subscribe, this function will time out and fail the test.
+	waitReadyChannelWithTimeout := func(t *testing.T, ready <-chan bool) (bool, bool) {
+		t.Helper()
+		select {
+		case actual, ok := <-ready:
+			return actual, ok
+		case <-time.After(100 * time.Millisecond):
+			t.Error("timeout waiting for unleash client to be ready after 100ms")
+		}
+		return false, false
+	}
+
+	t.Run("call ready", func(t *testing.T) {
+		assert := assert.New(t)
+		defer gock.OffAll()
+
+		gock.New(mockerServer).
+			Get("/client/features").
+			Reply(200).
+			JSON(api.FeatureResponse{})
+
+		mockListener := &MockedListener{}
+		mockListener.On("OnRegistered", mock.AnythingOfType("ClientData"))
+		mockListener.On("OnReady").Return()
+
+		client, err := NewClient(
+			WithUrl(mockerServer),
+			WithAppName(mockAppName),
+			WithInstanceId(mockInstanceId),
+			WithListener(mockListener),
+			WithDisableMetrics(true),
+		)
+		assert.NoError(err)
+
+		ready, ok := waitReadyChannelWithTimeout(t, client.Ready())
+		assert.True(ready, "the ready should be true when client has synced the feature flags")
+		assert.True(ok)
+
+		err = client.Close()
+
+		assert.True(gock.IsDone(), "there should be no more mocks")
+	})
+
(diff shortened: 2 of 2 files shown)


FredrikOseberg:
@kitimark Thanks for the PR. I'm currently looking at it. So if I understand you correctly you are calling client.Ready() multiple times and the second time you call it it will hang because the channel is closed?

youssefkhdessouky:
@FredrikOseberg I encountered this issue during development when working with unleash-client-go. Specifically, when calling unleash.WaitForReady(), the application becomes stuck if the Unleash URL is incorrect or the Unleash server/API is unavailable. This happens because WaitForReady() blocks until the client successfully connects and retrieves the initial feature toggle state, and in cases where the connection cannot be established, it never proceeds.

FredrikOseberg:
> @FredrikOseberg I encountered this issue during development when working with unleash-client-go. Specifically, when calling unleash.WaitForReady(), the application becomes stuck if the Unleash URL is incorrect or the Unleash server/API is unavailable. This happens because WaitForReady() blocks until the client successfully connects and retrieves the initial feature toggle state, and in cases where the connection cannot be established, it never proceeds.

This PR does not address WaitForReady though. And that would be the intended behavior of WaitForReady, in case you have feature flags that must present before the application starts. You should be able to have WaitForReady resolve by reading a backup file on disk or providing bootstrapped values. It doesn't need to depend on a network call to resolve, it just needs to have feature flags in the repository and ready to serve.

kitimark:
> @kitimark Thanks for the PR. I'm currently looking at it. So if I understand you correctly you are calling client.Ready() multiple times and the second time you call it it will hang because the channel is closed?

In our failing case we only call client.Ready() once. The problem shows up when the client finishes bootstrapping before we even start listening to the channel. The root cause is the repo signals readiness straight into uc.ready, the sync loop consumes that value immediately, so our later <-client.Ready() call just blocks. We wrap it in a timeout to guard against bad URLs or flaky networks, and that timeout is what trips, even though the client is already ready.

The reproduce code in the [Issue](https://github.com/Unleash/unleash-go-sdk/issues/203) forces that ordering so it happens every time, but it mirrors what we’ve observed in practice when the network responds quickly.

The proposed fix is **_workaround_** for makes Ready() behave like the docs suggest: each call ties into the onReady latch, so it delivers instantly if we subscribe after the client is ready, and multiple sources signalling readiness remain harmless.

Longer-term we should consider about uc.ready from this discussion point on this pull request description

> Currently repository (and in streaming, the delta processor) can signal readiness into uc.ready more than once. Closing uc.onReady once makes duplicate sources harmless, but confirm this is the intended single “ready” moment across modes.

If this uc.ready abstraction is correct. we should fix this instead to make it signal readiness at once and handle it correct following the client.Ready is provided.

PR #208, opened 2025-11-07
chore: don't emit metrics event when it isn't needed

sighphyre:
Currently merging into the wrong branch so we can see a nice diff.

This is a little hack to not emit metrics events when they're not actually listened to. Improves the performance under load for p50 -> p99 by about 2 orders of magnitude


sighphyre:
This was a proof of concept. Closing this now so we can do this properly

PR #209, opened 2025-11-10
Chore/remove build target

FredrikOseberg:
Removes 1.21 as a build target now that we are using math/rand/v2. Adds 1.25.


chriswk (approved):
LGTM

PR #210, opened 2025-11-18
fix: makes the storage layer return pointers

sighphyre:
The storage layer now returns pointers to the underlying data, rather than values. This fixes some aggressive GC thrashing that occurred on every feature evaluation when under load


storage.go
```diff
@@ -59,42 +59,54 @@ func (ds *DefaultStorage) Reset(data map[string]interface{}, persist bool) error
...
+	if err != nil {
 		return err
-	} else {
-		dec := json.NewDecoder(file)
-		var featuresFromFile map[string]api.Feature
-		if err := dec.Decode(&featuresFromFile); err != nil {
-			return err
-		}
+	}
+	defer file.Close()
```
sighphyre, line 37: Added this, since we previously leaked the file handle here. Generally not a problem on 'nix systems but in Windows this typically means the file can't be written to until the process exits. We probably never noticed because we don't actually log backup failures

bootstrap_storage.go
```diff
@@ -38,18 +38,18 @@ func (bs *BootstrapStorage) Init(backupPath string, appName string) {
...
 func (bs *BootstrapStorage) Persist() error {
 	return bs.backingStore.Persist()
 }
 
-func (bs *BootstrapStorage) Get(key string) (interface{}, bool) {
+func (bs *BootstrapStorage) Get(key string) (*api.Feature, bool) {
 	return bs.backingStore.Get(key)
 }
 
-func (bs *BootstrapStorage) List() []interface{} {
```
FredrikOseberg: This can't go into v5 as it's a breaking change

FredrikOseberg (changes requested):
Isolated changes look fine, but we can't merge this into current v5 branch. Needs to go into v6.

bootstrap_storage.go
sighphyre: Yep, pointed it at v6 now

FredrikOseberg (approved):

PR #211, opened 2025-11-18
chore: public api works on atomic pointers

sighphyre:
First step to making this lock free is to stop using patterns that require locking. This PR makes some overdue changes to the way data is held in the SDK

- The Datastore layer is now a data store. It no longer caches, that's now handled elsewhere. This has the nice side effect of allowing caching to work correctly when end users implement their own caching
- Segments are now correctly persisted, meaning first load from disk with the SDK will no longer cause features using segments to fail until API hydration kicks in
- Segments and features are no longer cached independently in different layers in the SDK
- Segments and features are held together behind an atomic pointer, meaning synchronization is now done by hardware and not between the Go runtime and the OS
- Feature evaluation logic is broken into it's own home, expressed as a set of functions that operate on the held feature state

Overall this makes the p99.9 unmeasurably small when operating on a standard strategy feature (which in theory is just a boolean return).


bootstrap_storage.go
```diff
@@ -13,43 +12,31 @@ type BootstrapStorage struct {
...
 }
 
-func (bs *BootstrapStorage) Load() error {
-	if len(bs.backingStore.data) > 0 || bs.Reader == nil {
-		return nil
+func (ds *BootstrapStorage) Init(backupPath, appName string) {
+	ds.backingStore.Init(backupPath, appName)
+}
+
+func (bs *BootstrapStorage) Load() (*api.FeatureResponse, error) {
```
sighphyre, line 19: The logic here needs to change because we now only access that data store when storing or retrieving data. So now we no longer early exit when failing to read the bootstrap, instead we attempt to fallback to the store and if that fails only then do we return an error. In practice I'm not sure the error handling matters much here since we silently ignore the error anyway

bootstrap_storage.go
```diff
@@ -13,43 +12,31 @@ type BootstrapStorage struct {
...
 
-func (bs *BootstrapStorage) Reset(data map[string]*api.Feature, persist bool) error {
-	return bs.backingStore.Reset(data, persist)
-}
-
-func (bs *BootstrapStorage) Persist() error {
-	return bs.backingStore.Persist()
-}
-
-func (bs *BootstrapStorage) Get(key string) (*api.Feature, bool) {
```
sighphyre, line 49: Get isn't needed anymore since this isn't the cache now. Instead the equivalent responsibility is done on the repository layer

bootstrap_storage.go
```diff
@@ -13,43 +12,31 @@ type BootstrapStorage struct {
...
-	return bs.backingStore.Get(key)
-}
-
-func (bs *BootstrapStorage) List() []*api.Feature {
```
sighphyre, line 53: Likewise this function no longer makes sense. Honestly it probably never made sense on the bootstrapping layer

client.go
```diff
@@ -283,7 +282,8 @@ func (uc *Client) sync() {
 //
 // It is safe to call this method from multiple goroutines concurrently.
 func (uc *Client) IsEnabled(feature string, options ...FeatureOption) (enabled bool) {
-	result, f := uc.isEnabled(feature, options...)
+	snapshot := uc.repository.snapshot()
```
sighphyre, line 285: This is effectively the magic. The repository can yield up snapshots of it's current state of features and segments. That snapshot is backed by an atomic pointer, meaning it remains valid for as long as the pointer is held, while the repository is allowed to swap its own pointer for a new value without affecting requests in flight

client.go
```diff
@@ -313,13 +313,17 @@ func (uc *Client) IsEnabled(feature string, options ...FeatureOption) (enabled b
 
 // isEnabled abstracts away the details of checking if a toggle is turned on or off
 // without metrics
-func (uc *Client) isEnabled(feature string, options ...FeatureOption) (api.StrategyResult, *api.Feature) {
+func (uc *Client) isEnabled(feature string, snapshot *FeatureMemoryState, options ...FeatureOption) (api.StrategyResult, *api.Feature) {
```
sighphyre, line 316: We must operate on the same pointer, otherwise we risk tearing the evaluation. That is actually a substantial risk on the existing implementation - that a single feature evaluation may be operating on a feature or segment that changes during the course of the evaluation

client.go
```diff
@@ -626,3 +523,11 @@ func handleFallback(opts featureOption, featureName string, ctx *context.Context
 		Enabled: false,
 	}
 }
+
+func resolveToggle(snapshot *FeatureMemoryState, opts featureOption, featureName string) *api.Feature {
```
sighphyre, line 527: This needs to be removed. This is a work around because the way we designed our storage interface means that end users implementing their own would lose access to caching. However, this is in the public API and it's hard and painful to remove now and this needs to be a different PR's problem

delta_processor.go
```diff
@@ -8,23 +8,22 @@ import (
 )
 
 type deltaProcessor struct {
-	storage            Storage
-	repository         *repository // Repository reference for segment manipulation
+	repository         *repository // ideally this shouldn't be necessary, but for now we're going going to use it to resolve a snapshot of the feature state
```
sighphyre, line 11: In the spirit of keeping this PR self contained, I've made the smallest possible changes to this file. That being said, this needs to be dealt with soonish. The circular dependency on repository is going to cause problems and there's a lot of inappropriate intimacy with other modules. Ideally this becomes a family of pure functions which collapse deltas into a feature response and allow other modules to deal with the other work

delta_processor.go
```diff
@@ -79,10 +77,28 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
 		}
 	}
 
-	if err := dp.repository.updateStorageWithDelta(currentFeatures, segments); err != nil {
```
sighphyre, line 82: I've chosen to discard this error handling branch because I think its wrong. This can only occur if writing to the persistent storage fails in the v5 implementation. The data store is still updated, which means feature flags are still working correctly. I think this is especially problematic with custom implementations of the data store where we don't control the error paths

feature_state.go
```diff
@@ -0,0 +1,163 @@
+package unleash
```
sighphyre, line 1: Harvested API from Client. This is effectively a state that exposes methods that determine if a given feature/variant is enabled for that state. This isn't quite there yet in terms of architecture but it's closer to how this conceptually should work. Don't want to do more here because this PR is already quite large

feature_state.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/Unleash/unleash-go-sdk/v5/strategy"
+)
+
+type FeatureMemoryState struct {
+	Features map[string]*api.Feature
+	Segments map[int][]api.Constraint
+}
+
+// evaluateFeature applies strategies + constraints for a single feature.
+// It does NOT handle fallbacks or missing features; that's the caller's job.
```
sighphyre, line 19: Similar to how other SDKs handle this. Top level APIs deal with fallbacks, missing features + metrics. Engine layers deal with computational evaluation of a flag

repository.go
```diff
@@ -66,22 +66,44 @@ func newRepository(options repositoryOptions, channels repositoryChannels) *repo
...
-		)
+			repo.deltaProcessor)
 	}
 
 	go repo.sync()
 
 	return repo
 }
 
+func (r *repository) updateState(features *api.FeatureResponse) {
```
sighphyre, line 93: One place to control access to mutating the state. This isn't perfect and should be its own thing. But that requires some redesign of the repository layer

repository.go
```diff
@@ -136,9 +158,6 @@ func (r *repository) sync() {
 			if r.streamingClient != nil {
 				r.streamingClient.stop()
 			}
-			if err := r.options.storage.Persist(); err != nil {
```
sighphyre, line 139: I actually don't know what this is intended to achieve. So I've done the dumbest thing possible and removed it. Neither an error case nor a close case should cause a persistence call. Persist is now done explicitly where it's needed - when new data is fetched. Once via delta processing and once via old style feature fetch

repository.go
```diff
@@ -223,29 +242,14 @@ func (r *repository) fetch() error {
...
-	r.segments = featureResp.SegmentsMap()
-	r.options.storage.Reset(featureResp.FeatureMap(), true)
+	r.saveState(&featureResp)
 	r.successfulFetch()
 	r.Unlock()
 	return nil
 }
 
-// updateStorageWithDelta updates the storage with delta changes in a thread-safe manner
-func (r *repository) updateStorageWithDelta(features map[string]*api.Feature, segments map[int][]api.Constraint) error {
```
sighphyre, line 237: This now goes via the one and only one way to update state. I think this method only really exists because the delta processor and the repository are strangely coupled at the moment

repository.go
```diff
@@ -223,29 +242,14 @@ func (r *repository) fetch() error {
...
-	r.segments = segments
-
-	// Update storage
-	return r.options.storage.Reset(features, true)
-}
-
 // IsStreaming returns whether the repository is currently in streaming mode
 func (r *repository) IsStreaming() bool {
-	r.RLock()
-	defer r.RUnlock()
```
sighphyre, line 251: The streaming property is set on construction and never mutated. This lock does nothing but introduce contention that doesn't need to be here

repository.go
```diff
@@ -264,38 +268,10 @@ func (r *repository) statusIsOK(resp *http.Response) error {
...
 
-func (r *repository) getToggle(key string) *api.Feature {
-	r.RLock()
-	defer r.RUnlock()
-
-	feature, _ := r.options.storage.Get(key)
-	return feature
-}
-
-func (r *repository) resolveSegmentConstraints(strategy api.Strategy) ([]api.Constraint, error) {
```
sighphyre, line 278: This absolutely does not belong in this layer. It's been moved to feature_state where it can be happy and lock free

storage.go
```diff
@@ -9,106 +9,52 @@ import (
 	"github.com/Unleash/unleash-go-sdk/v5/api"
 )
 
-// Storage is an interface that can be implemented in order to have control over how
-// the repository of feature toggles is persisted.
+// Storage controls persistence of the SDK state (features + segments).
```
sighphyre, line 12: This module is the heart of a lot of the problems we have. I've taken a knife to it. It now does what it says on the tin. It stores stuff. There's more to be done here but in another PR

**This unfortunately is a breaking change**

streaming_client_test.go
```diff
@@ -22,6 +22,18 @@ func (m *mockEvent) Id() string    { return m.id }
 func (m *mockEvent) Event() string { return m.event }
 func (m *mockEvent) Data() string  { return m.data }
 
+type NoOpStorage struct{}
```
sighphyre, line 25: So that our tests are not loading from disk and therefore hydrating their initial state from the runs of other tests. It's not technically necessary for this PR but I've had a ton of problems in SDKs that don't correctly enforce a pattern for no op storage and it's substantially worse in complex integration tests

FredrikOseberg (approved):
I think this looks sane, but it's a big change and requires heavy testing. Maybe we can go over it tomorrow together?

PR #212, opened 2025-11-18
fix: use sync pool for strategy randomness

sighphyre:
Instead of blocking on a single shared Mutex, randomness now uses a sync pool, which gives multiple goroutines a bit more breathing room to get access to randomness without lining up one behind another

### DIFF
--- a/api/feature.go
+++ b/api/feature.go
@@ -86,12 +86,12 @@ func (fr FeatureResponse) SegmentsMap() map[int][]Constraint {
 }
 
 // Get variant for a given feature which is considered as enabled
-func (vc VariantCollection) GetVariant(ctx *context.Context) *Variant {
+func (vc VariantCollection) GetVariant(ctx *context.Context, stickinessOverride *string) *Variant {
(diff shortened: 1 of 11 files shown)


FredrikOseberg (approved):
This looks sane. Nice! Also this is backwards compatible so we can merge it towards both branches.

EDIT: Is the target branch here correct Simon? I would imagine you want to merge this into v5 or v6.

sighphyre:
> This looks sane. Nice! Also this is backwards compatible so we can merge it towards both branches.
> 
> EDIT: Is the target branch here correct Simon? I would imagine you want to merge this into v5 or v6.

Yessir, it's correct. This is a stacked PR, if you're not familiar with the flow, it's worth doing some reading on it because it's a hella useful flow for this kind of situation. This entire chain builds on top of itself so this entire set of PRs require the work before them. Stacking the PR lets me work without being blocked by review and prevents you from having to review a 4k +/- monstrosity

The base now points to v6, so as these merge they'll all fall into that

PR #213, opened 2025-11-19
chore: use less locking in metrics hot paths

sighphyre:
Yeets the locking in the metrics code in favor of sync primitives. Fixes a bug in the re-insertion code since I was in that space. Variants are still behind a mutex but that mutex is now smaller scoped and locked less often. We can probably remove that too but its a bigger PR and I want to keep this focused on the hot path

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/Unleash/unleash-go-sdk/v5/internal/api"
@@ -87,20 +88,28 @@ type metric struct {
 	Enabled bool
 }
 
+type toggleCounters struct {
+	yes int64
+	no  int64
+
+	mu       sync.Mutex
+	variants map[string]int64
+}
+
 type metrics struct {
 	metricsChannels
-	options  metricsOptions
-	started  time.Time
-	bucketMu sync.Mutex
-	bucket   api.Bucket
-	ticker   *time.Ticker
-	close    chan struct{}
-	closed   chan struct{}
-	ctx      context.Context
-	cancel   func()
-	maxSkips float64
-	errors   float64
-	skips    float64
+	options         metricsOptions
+	started         time.Time
+	last_close_time time.Time
+	counters        sync.Map // map[string]*toggleCounters
+	ticker          *time.Ticker
+	close           chan struct{}
+	closed          chan struct{}
+	ctx             context.Context
+	cancel          func()
+	maxSkips        float64
+	errors          float64
+	skips           float64
 }
 
 func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
@@ -113,6 +122,7 @@ func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
 		maxSkips:        10,
 		errors:          0,
 		skips:           0,
+		last_close_time: time.Now(),
 	}
 	ctx, cancel := context.WithCancel(context.Background())
 	m.ctx = ctx
@@ -122,7 +132,6 @@ func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
 		m.options.httpClient = http.DefaultClient
 	}
 
-	m.resetBucket()
 	if m.options.metricsInterval <= 0 {
 		m.options.disableMetrics = true
 	}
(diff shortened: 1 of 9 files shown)


metrics.go
```diff
@@ -274,24 +336,57 @@ func (m *metrics) doPost(url *url.URL, payload interface{}) (*http.Response, err
...
+	}
+
+	nc := &toggleCounters{
+		variants: make(map[string]int64),
+	}
+	actual, _ := m.counters.LoadOrStore(name, nc)
+	return actual.(*toggleCounters)
+}
+
+func (m *metrics) reinsertBucket(bucket api.Bucket) {
```
sighphyre, line 352: Previously the code just dropped variants on the floor when re-inserting. Needed to get fixed and now seems like a good time I guess so here's a fix

metrics.go
```diff
@@ -274,24 +336,57 @@ func (m *metrics) doPost(url *url.URL, payload interface{}) (*http.Response, err
...
-			Variants: map[string]int32{},
-		}
-	}
+	c := m.getOrCreateCounter(name)
 	if enabled {
-		t.Yes += num
+		atomic.AddInt64(&c.yes, int64(num))
 	} else {
-		t.No += num
+		atomic.AddInt64(&c.no, int64(num))
```
sighphyre, line 386: Wheee atomic primitives are awesome. We can do this on the hardware level and not sandwiched between the runtime and OS

metrics_test.go
```diff
@@ -592,3 +599,36 @@ func TestMetrics_metricsData_includes_new_metadata(t *testing.T) {
 
 	st.Expect(t, gock.IsDone(), true)
 }
+
+func TestReinsertingBucketsAlsoRestoresVariants(t *testing.T) {
```
sighphyre, line 606: New test for the re-insert variant logic

metrics.go
```diff
@@ -87,20 +88,28 @@ type metric struct {
...
-	close    chan struct{}
-	closed   chan struct{}
-	ctx      context.Context
-	cancel   func()
-	maxSkips float64
-	errors   float64
-	skips    float64
+	options         metricsOptions
+	started         time.Time
+	last_close_time time.Time
```
FredrikOseberg: snake_case break camelCase convention

metrics.go
```diff
@@ -113,6 +122,7 @@ func newMetrics(options metricsOptions, channels metricsChannels) *metrics {
 		maxSkips:        10,
 		errors:          0,
 		skips:           0,
+		last_close_time: time.Now(),
```
FredrikOseberg: See above

metrics.go
```diff
@@ -196,13 +205,70 @@ func (m *metrics) successfulPost() {
...
+		if yes == 0 && no == 0 {
+			c.mu.Lock()
+			emptyVariants := len(c.variants) == 0
+			c.mu.Unlock()
+			if emptyVariants {
+				return true
+			}
+		}
+
+		tc := api.ToggleCount{
```
FredrikOseberg: toggleCount would be clearer

metrics.go
```diff
@@ -196,13 +205,70 @@ func (m *metrics) successfulPost() {
...
+// little bit more memory is held than necessary
+func (m *metrics) buildBucketAndReset(last_close_time time.Time) (api.Bucket, bool) {
+	bucket := api.Bucket{
+		Start:   last_close_time,
+		Toggles: make(map[string]api.ToggleCount),
+	}
+
+	m.counters.Range(func(key, value any) bool {
+		name := key.(string)
+		c := value.(*toggleCounters)
```
FredrikOseberg: counters? I find myself scrolling up to figure out what this variable is.

metrics.go
```diff
@@ -274,24 +336,57 @@ func (m *metrics) doPost(url *url.URL, payload any) (*http.Response, error) {
...
+	nc := &toggleCounters{
+		variants: make(map[string]int64),
+	}
+	actual, _ := m.counters.LoadOrStore(name, nc)
+	return actual.(*toggleCounters)
+}
+
+func (m *metrics) reinsertBucket(bucket api.Bucket) {
+	for name, tc := range bucket.Toggles {
+		c := m.getOrCreateCounter(name)
```
FredrikOseberg: counter?

FredrikOseberg (approved):
Some nits, otherwise LGTM

metrics.go
sighphyre: Absolutely correct. My bad. Too much Rust lately. Moved to `lastCloseTime`

metrics.go
sighphyre: Fixed

metrics.go
sighphyre: Yep, took that suggestion. Uh... is there a pattern for this? The default Go community stance seems to be make variables as short and difficult to read as possible so I've accidentally followed that sometimes but I hate it. Is the a rule where we shouldn't do that?

metrics.go
sighphyre: Fair, patched

metrics.go
sighphyre: Also fair, also patched

PR #214, opened 2025-11-19
chore: make metric count channels best effort send

sighphyre:
Last major piece of the performance puzzle. This means feature evaluation events will get dropped under high load, but that's appropriate for an eventing system anyway


FredrikOseberg (approved):
LG, losing some counts seems like an acceptable tradeoff.

PR #215, opened 2025-11-20
fix: variant strategies inherit stickiness

sighphyre:
Fixes a bug in the variant strategy implementation. This fails the client spec tests roughly half the time so no idea why it hasn't been fixed yet. Can't fix this on v5 because it requires a change to a public API. This API should not be public but it is


FredrikOseberg (approved):
LG

PR #216, opened 2025-11-26
chore: bump go.mod to v6

daveleek:
This bumps the version in go.mod to v6 along with all references to v5


gastonfournier (approved):
LG!

PR #217, opened 2025-12-03
fix: global rand

FredrikOseberg:
Use math rand v2 instead of locking


sighphyre (approved):
Sure I think this makes sense. Looks safe, we can only try!

PR #218, opened 2025-12-03
perf: reduce allocations in hot path to fix P99.9 latency spikes

FredrikOseberg:
## Summary

Reduces memory allocations in the `IsEnabled()` hot path to eliminate P99.9 latency spikes caused by GC pressure under high concurrency.

**Problem:** Customers reported P99.9 latency spikes of ~50ms under high concurrency (500+ goroutines). Root cause was excessive allocations triggering frequent GC pauses.

**Changes:**
- Pool murmur3 hash objects using `sync.Pool` (reduces allocs from 3 to 2 per call)
- Eliminate string concatenation in hash computation (`groupId + ":" + id` → separate writes)
- Pre-allocate constraint slices with estimated capacity
- Simplify `randomString()` to use thread-safe `rand.IntN()` directly

## Benchmark results

Total calls: 2900000
Latency (min/mean/max): 291ns / 2.305µs / 20.236333ms
P50:    708ns
P95:    1.875µs 
P99:    11.167µs
P99.9:  34.916µs
P99.99: 5.641667ms

Spikes >1ms:   544 (0.019%)
Spikes >10ms:  110 (0.004%)
Spikes >50ms:  0 (0.000%)
Spikes >100ms: 0 (0.000%)

### DIFF
--- a/feature_state.go
+++ b/feature_state.go
@@ -156,7 +156,8 @@ func findStrategy(strats []strategy.Strategy, name string) strategy.Strategy {
 }
 
 func (featureState *FeatureMemoryState) resolveSegmentConstraints(strategy api.Strategy) ([]api.Constraint, error) {
-	segmentConstraints := []api.Constraint{}
+	// Pre-allocate with estimated capacity to avoid slice growth allocations
+	segmentConstraints := make([]api.Constraint, 0, len(strategy.Segments)*4)
 
 	segments := featureState.Segments
 
--- a/internal/strategies/helpers.go
+++ b/internal/strategies/helpers.go
@@ -13,18 +13,14 @@ import (
 
 var VariantNormalizationSeed uint32 = 86028157
 
-// colonByte is a pre-allocated byte slice to avoid allocation in hot path
 var colonByte = []byte(":")
 
-// Hash pools for the two seed values used in the SDK
-// Pool for seed=0 (used by normalizedRolloutValue)
 var hashPoolSeed0 = sync.Pool{
 	New: func() any {
 		return murmur3.SeedNew32(0)
 	},
 }
 
-// Pool for VariantNormalizationSeed (used by variant selection)
 var hashPoolVariantSeed = sync.Pool{
 	New: func() any {
 		return murmur3.SeedNew32(VariantNormalizationSeed)
@@ -66,7 +62,6 @@ func normalizedRolloutValue(id string, groupId string) uint32 {
 }
 
 func NormalizedVariantValue(id string, groupId string, normalizer int, seed uint32) uint32 {
-	// Use pooled hash objects for known seeds, fall back to allocation for unknown seeds
 	var h hash.Hash32
 	var pool *sync.Pool
 
@@ -78,7 +73,6 @@ func NormalizedVariantValue(id string, groupId string, normalizer int, seed uint
 		pool = &hashPoolVariantSeed
 		h = pool.Get().(hash.Hash32)
 	default:
-		// Fallback for any other seed (shouldn't happen in practice)
 		h = murmur3.SeedNew32(seed)
 	}
 
--- a/internal/strategies/helpers.go
+++ b/internal/strategies/helpers.go
@@ -1,6 +1,7 @@
 package strategies
 
 import (
+	"hash"
 	"math/rand/v2"
 	"os"
 	"strconv"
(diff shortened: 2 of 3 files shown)


feature_state.go
```diff
@@ -156,7 +156,8 @@ func findStrategy(strats []strategy.Strategy, name string) strategy.Strategy {
 }
 
 func (featureState *FeatureMemoryState) resolveSegmentConstraints(strategy api.Strategy) ([]api.Constraint, error) {
-	segmentConstraints := []api.Constraint{}
+	// Pre-allocate with estimated capacity to avoid slice growth allocations
+	segmentConstraints := make([]api.Constraint, 0, len(strategy.Segments)*4)
```
sighphyre, line 160: This seems a little suspect to me. I'd like to see some benchmarks actually stressing this before we make random changes. I think real world impacts of this are difficult to predict but this change against a toggle with no segments is net negative. I also don't think this is the correct way to tackle this problem, because we're allocating arrays at the call site anyway

internal/strategies/helpers.go
```diff
@@ -12,6 +13,20 @@ import (
 
 var VariantNormalizationSeed uint32 = 86028157
 
+var colonByte = []byte(":")
+
+var hashPoolSeed0 = sync.Pool{
```
sighphyre, line 18: I believe this is used in rollout strategies, can we call it that rather than `hashPoolSeed0`? I had to go and look why we need a 0 seed

internal/strategies/helpers.go
```diff
@@ -47,9 +62,32 @@ func normalizedRolloutValue(id string, groupId string) uint32 {
...
+	}
+
+	h.Reset()
+	// Write components separately to avoid string concatenation allocation
+	h.Write([]byte(groupId))
+	h.Write(colonByte)
+	h.Write([]byte(id))
+	hashCode := h.Sum32()
+
+	// Return to pool if we got it from there
```
sighphyre, line 86: Personal taste but I'd rather we clean up AI comments that add nothing to the reading of the code. The one above here is fine but this is noise

```suggestion
```

internal/strategies/helpers_test.go
```diff
@@ -77,13 +77,29 @@ func TestNewRng(t *testing.T) {
...
+
+	goRoutines := 20
+	wg.Add(goRoutines)
+	for range goRoutines {
+		go testGen(100)
+	}
+	wg.Wait()
+}
+
+func TestRandomString(t *testing.T) {
```
sighphyre, line 96: I don't think this test does anything useful for us. `randomString` is now a direct wrapper around `randv2`. If there's an important case encoded in this test I think we should give it a proper name that informs the reader what that case is

internal/strategies/helpers_test.go
```diff
@@ -77,13 +77,29 @@ func TestNewRng(t *testing.T) {
 	testGen := func(n int) {
 		for range n {
 			randomInt := rng.int()
-			assert.True(t, randomInt >= 0 && randomInt <= 100)
-
-			randomString := rng.string()
-			assert.True(t, len(randomString) <= 5)
+			assert.True(t, randomInt >= 1 && randomInt <= 101)
```
sighphyre, line 80: This doesn't look right to me. Is there a reason we're shifting the numeric value by one here? If this changes unintentionally then I think we've broken something, if this test doesn't need to change then we should revert this

sighphyre (commented):
I'd really like to split this into two separate PRs if possible. It's really difficult to know which of of the changes here is helping and which isn't (I suspect both are but what impacts they're having yields insights on where to go next). Also be nice to include both before and after benchmarks on the results here, post benchmarks in isolation aren't hugely helpful. Also, would you mind giving this description a human once over? It's very AI gen at the moment and I don't think it's completely and correctly capturing either our current understanding of the problem or the full details around the changes this makes

FredrikOseberg:
> I'd really like to split this into two separate PRs if possible. It's really difficult to know which of of the changes here is helping and which isn't (I suspect both are but what impacts they're having yields insights on where to go next). Also be nice to include both before and after benchmarks on the results here, post benchmarks in isolation aren't hugely helpful. Also, would you mind giving this description a human once over? It's very AI gen at the moment and I don't think it's completely and correctly capturing either our current understanding of the problem or the full details around the changes this makes

Happy to massage this later. I set this up end of day today so you could have a quick look.

sighphyre:
I think we ended up covering this through other work. I'm going to close this @FredrikOseberg, if you want it, feel free to reopen!

PR #221, opened 2026-01-16
chore: rework public api to reduce allocations

sighphyre:
This reworks the public API to not use a functional interface to pass options. That pattern is ergonomically nice but requires that the compiler do a bunch of moves of that data as we compute the actual values. Turns out that's super expensive.

Not convinced this is the API that we want. I am open to suggestions. Please! 

Combined with #220, this massively reduces the allocations on the isEnabled hot path.

Size allocations after this and #220 - isEnabled no longer appears here :smile:  :

There's potentially more we can do here but I think a reduction of several gigabytes to unnoticeable is probably good enough for this pass

Type: alloc_space
Time: 2026-01-16 16:29:57 SAST
Showing nodes accounting for 18262.89kB, 100% of 18262.89kB total
      flat  flat%   sum%        cum   cum%
    3591kB 19.66% 19.66%     3591kB 19.66%  runtime.allocm
 2048.16kB 11.21% 30.88%  2048.16kB 11.21%  crypto/x509.parseName
 1974.97kB 10.81% 41.69%  1974.97kB 10.81%  bytes.growSlice
 1928.99kB 10.56% 52.25%  1928.99kB 10.56%  reflect.growslice
 1025.50kB  5.62% 57.87%  5634.09kB 30.85%  crypto/x509.parseCertificate
 1025.31kB  5.61% 63.48%  1025.31kB  5.61%  encoding/pem.Decode
 1024.09kB  5.61% 69.09%  1024.09kB  5.61%  crypto/x509/pkix.(*Name).FillFromRDNSequence
 1000.34kB  5.48% 74.57%  2757.67kB 15.10%  main.main
  544.67kB  2.98% 77.55%   544.67kB  2.98%  net.open
  515.19kB  2.82% 80.37%   515.19kB  2.82%  unicode.map.init.2
  512.28kB  2.81% 83.18%   512.28kB  2.81%  math/big.nat.make (inline)
  512.22kB  2.80% 85.98%   512.22kB  2.80%  runtime.malg
  512.09kB  2.80% 88.79%   512.09kB  2.80%  github.com/Unleash/unleash-go-sdk/v5/api.FeatureResponse.FeatureMap (inline)
  512.03kB  2.80% 91.59%   512.03kB  2.80%  vendor/golang.org/x/crypto/cryptobyte.(*String).ReadASN1ObjectIdentifier
  512.02kB  2.80% 94.39%   512.02kB  2.80%  crypto/x509.processExtensions
[...]

PR #222, opened 2026-01-19
chore: clean up public api

sighphyre:
Drops the old IsEnabled functions that required the functional interface and swapped in IsEnabledWithOptions.

I don't really want to support the old methods if they're going to cause so many performance problems so we're doing a hard break here

Most of this is patching tests. The important work got did in #221

### DIFF
--- a/client.go
+++ b/client.go
@@ -337,7 +337,6 @@ func (uc *Client) IsEnabledWithOptions(feature string, options FeatureOptions) (
 	return
 }
 
-// IsEnabledWithOptions is the hot-path friendly API.
 func (uc *Client) isEnabledWithOptions(
 	feature string,
 	snapshot *FeatureMemoryState,
--- a/client.go
+++ b/client.go
@@ -321,9 +321,7 @@ func (uc *Client) IsEnabledWithOptions(feature string, options FeatureOptions) (
 
 		if f != nil && f.ImpressionData && uc.impressionListener != nil {
 			ctx := uc.staticContext
-			if options.HasCtx {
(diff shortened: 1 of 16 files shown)


client.go
```diff
@@ -643,14 +484,14 @@ func (uc *Client) ListFeatures() []api.Feature {
 	return uc.repository.list()
 }
 
-func handleFallback(opts featureOption, featureName string, ctx *context.Context) api.StrategyResult {
-	if opts.fallbackFunc != nil {
+func handleFallback(fallbackFunc FallbackFunc, fallback *bool, featureName string, ctx *context.Context) api.StrategyResult {
```
sighphyre, line 487: Small rework here to improve the sanity. This previously took in a featureOption, which had a context on it, which this function was absolutely not supposed to use because it hadn't been merged with the static context. This is harder to misuse at the cost of another parameter.

client.go
```diff
@@ -659,9 +500,9 @@ func handleFallback(opts featureOption, featureName string, ctx *context.Context
 	}
 }
 
-func resolveToggle(snapshot *FeatureMemoryState, opts featureOption, featureName string) *api.Feature {
-	if opts.resolver != nil {
-		return opts.resolver(featureName)
+func resolveToggle(snapshot *FeatureMemoryState, resolver FeatureResolver, featureName string) *api.Feature {
```
sighphyre, line 503: No need to ingest such a broad type when we can just accept what we need

FredrikOseberg (approved):
LG

PR #224, opened 2026-01-29
feat: shared metrics types for impact metrics

kwasniew:
## About the changes

Porting first shared types for impact metrics based on this: https://github.com/Unleash/unleash-node-sdk/blob/main/src/impact-metrics/metric-types.ts

Everything will be tested transitively once we add actual metrics (counter, gauge, histogram). This is extracted from 1000 LOC PR.

### DIFF
--- a/impactmetrics/metric_types.go
+++ b/impactmetrics/metric_types.go
@@ -1,0 +1,55 @@
+package impactmetrics
+
+import (
+	"sort"
+	"strings"
+)
+
+type MetricLabels map[string]string
+
+type NumericMetricSample struct {
+	Labels MetricLabels `json:"labels"`
+	Value  float64      `json:"value"`
+}
+
+type CollectedMetric struct {
+	Name    string        `json:"name"`
+	Help    string        `json:"help"`
+	Type    string        `json:"type"`
+	Samples []interface{} `json:"samples"`
+}
+
+type ImpactMetricsDataSource interface {
+	Collect() []CollectedMetric
+	Restore(metrics []CollectedMetric)
+}
+
+func LabelKey(labels MetricLabels) string {
+	if len(labels) == 0 {
+		return ""
+	}
+	keys := make([]string, 0, len(labels))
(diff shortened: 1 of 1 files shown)


impactmetrics/metric_types.go
```diff
@@ -0,0 +1,55 @@
...
+type NumericMetricSample struct {
+	Labels MetricLabels `json:"labels"`
+	Value  float64      `json:"value"`
+}
+
+type CollectedMetric struct {
+	Name    string        `json:"name"`
+	Help    string        `json:"help"`
+	Type    string        `json:"type"`
+	Samples []interface{} `json:"samples"`
```
sjaanus, line 19: Can we add types for Samples. Currently if I understand it correctly, its not strongly typed.

impactmetrics/metric_types.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return strings.Join(parts, ",")
+}
+
+func ParseLabelKey(key string) MetricLabels {
+	labels := MetricLabels{}
+	if key == "" {
+		return labels
+	}
+	for _, pair := range strings.Split(key, ",") {
```
sjaanus, line 48: This i probably in every language, but just thinking about it. Can we have a label that has comma inside? What breaks?

sjaanus (approved):
LG with comments

impactmetrics/metric_types.go
kwasniew: The tradeoff is: stronger typing now means more JSON serialization complexity (when we add histograms). The current []interface{} with json tags on the concrete structs "just works" for marshaling

FredrikOseberg (approved):
LG

impactmetrics/metric_types.go
kwasniew: good question. I think it's a limitation across all sdks but we'd need appName or environment with coma. We would drop the part after the comma.

PR #225, opened 2026-01-30
feat: impact metrics counter

kwasniew:
## About the changes

Porting counter implementation from https://github.com/Unleash/unleash-node-sdk/blob/main/src/impact-metrics/metric-types.ts 
Tries to keep this PR as small as possible. There's only 100 LOC for prod impl. The rest is verbose tests

Important difference from Node SDK:
* trade safety with mutexes

Design decisions in inline comments

## Discussion points

After PR review applied the following changes:
* avoid casting int to float in collect/restore cycle
* removed unnecessary counterimpl casting
* removed insertion order keeping and sorting keys for easier testing instead
* moved all the code to internal for now, will start exposing public API when we get to it
* handling negative counter values by dropping them (test added)


impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+type Counter interface {
+	Inc(value int64, labels MetricLabels)
+}
+
+type counterImpl struct {
+	mu     sync.Mutex
+	name   string
+	help   string
+	values map[string]int64
+	keys   []string // insertion order
```
kwasniew: Node preserves insertion order but go doesn't so we need to help it

internal/impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+
+type Counter interface {
+	Inc(value int64, labels MetricLabels)
+}
+
+type counterImpl struct {
+	mu     sync.Mutex
+	name   string
+	help   string
+	values map[string]int64
```
kwasniew, line 66: internally we use int64 but convert to float64 to share NumericMetricSample wire format with gauges

internal/impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+
+func newCounter(name, help string) *counterImpl {
+	return &counterImpl{
+		name:   name,
+		help:   help,
+		values: map[string]int64{},
+	}
+}
+
+func (c *counterImpl) Inc(value int64, labels MetricLabels) {
```
kwasniew, line 77: requires explicit value and label. No optional params as Node. The use-facing MetricAPI will provide overrides with different variants but internal type is kept simple

internal/impactmetrics/metric_types_test.go
```diff
@@ -0,0 +1,155 @@
+package impactmetrics
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCounterIncrementsByDefaultValue(t *testing.T) {
```
kwasniew, line 9: all tests ported from node sdk tests 1 to 1

impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+
+func (c *counterImpl) collect() CollectedMetric {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	samples := make([]interface{}, 0, len(c.values))
+	for _, key := range c.keys {
+		samples = append(samples, NumericMetricSample{
+			Labels: ParseLabelKey(key),
+			Value:  float64(c.values[key]),
```
kwasniew: satisfy wire format shared with gauges

impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+	return CollectedMetric{
+		Name:    c.name,
+		Help:    c.help,
+		Type:    "counter",
+		Samples: samples,
+	}
+}
+
+type InMemoryMetricRegistry struct {
+	mu          sync.RWMutex
```
kwasniew: we don't need full write low for read operations

impactmetrics/metric_types.go
sighphyre: Oh... uh... is that a requirement? I don't remember seeing it in Ygg, Java or the spec. Do we need to fix those?

impactmetrics/metric_types.go
sighphyre: Personal feeling but I think this is the worst layer to be doing the casting on. This means that we have to bounce back and forth between floats and integers when we collect/restore. I think either accepting an integer at the public boundary and immediately converting it or defering this until send would both be better options

internal/impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	key := LabelKey(labels)
```
sighphyre, line 83: Is this a hotpath? I don't think this is a hot path but I'm seeing comments and patterns that suggest it is. If this is a hot path, LabelKey is going to take the GC out back and beat it with a stick

impactmetrics/metric_types.go
sighphyre: RWMutexs do best when reads are non trivial. I think a sync Map would probably out perform this substantially (100% gut feel, and whether or not we look depends on how much we care about perf here). There's a good example on the v6 branch around metrics that should guide you if you're interested

But also, this won't matter until we fix the label keying because that is almost certainly going to dominate here

impactmetrics/metric_types.go
```diff
@@ -53,3 +54,131 @@ func ParseLabelKey(key string) MetricLabels {
...
+		return []CollectedMetric{}
+	}
+	return result
+}
+
+func (r *InMemoryMetricRegistry) Restore(metrics []CollectedMetric) {
+	for _, m := range metrics {
+		switch m.Type {
+		case "counter":
+			c, ok := r.Counter(m.Name, m.Help).(*counterImpl)
```
sighphyre: The cast here is very unfortunate. This pattern works fine in TS/Rust but its not great in Go. I kinda wanna suggesting moving delegating this behavior to the metric itself if we can

sighphyre (commented):
This is mostly internal stuff. This definitely, definitely cannot go into this folder. This belongs in  /internal where the compiler can prevent it leaking out. Otherwise we're just throwing everything here into the public API

impactmetrics/metric_types.go
kwasniew: i wanted to have identical unit tests between Node and Golang SDKs. It was a local SDK decision. I don't think it's important to specify this in a global spec for now. I can drop this one and have a slightly different test in Go with sorting by key strings rather than insertion order.

kwasniew: removed

impactmetrics/metric_types.go
kwasniew: introduced IntMetricSample for now and the casting is not needed for now. I will see how it evolves with other metric types

impactmetrics/metric_types.go
kwasniew: removed the cast. it was a mistake to have it in the first place. no need to delegate it to metrics yet. I will observe what feedback I get from the code with the next metrics

internal/impactmetrics/metric_types.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (c *counterImpl) collect() CollectedMetric {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	keys := make([]string, 0, len(c.values))
+	for k := range c.values {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
kwasniew, line 95: negligible perf impact but gives predictable testing output

sighphyre: As soon as we start to put test specific details in prod code, we chain the implementation to those details, even though they're not requirements. That makes someone's life coming into this without context much harder than it needs to be.

I'd really like to not do things like this but if you're married to it, can we at least leave a comment for someone in future, because these are the kind of details that send me down rabbitholes later when I'm trying to understand why a given SDK does what it does

internal/impactmetrics/metric_types.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (r *InMemoryMetricRegistry) Collect() []CollectedMetric {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	names := make([]string, 0, len(r.counters))
+	for name := range r.counters {
+		names = append(names, name)
+	}
+	sort.Strings(names)
```
sighphyre, line 161: Same as above. I can't tell if this is a requirement or a detail where we've entangled test and prod paths. If it's the latter, breadcrumbs for the next person in the form of a comment would be amazing

sighphyre (approved):
Thank you for moving this!

internal/impactmetrics/metric_types.go
kwasniew: I can explore one other option and sort keys in the test itself. will give it a try.

internal/impactmetrics/metric_types.go
kwasniew: will address in the very next PR

PR #226, opened 2026-01-30
feat: impact metrics gauge

kwasniew:
## About the changes

Adds gauges similar to how we added counters before. This PR documents the sorting of sample keys. We may revisit this decision since spec doesn't specify this part.

### DIFF
--- a/internal/impactmetrics/metric_types.go
+++ b/internal/impactmetrics/metric_types.go
@@ -43,6 +43,7 @@ func LabelKey(labels MetricLabels) string {
 	for k := range labels {
 		keys = append(keys, k)
 	}
+	// Sort keys for deterministic output since Go map iteration order is random
 	sort.Strings(keys)
 	parts := make([]string, len(keys))
 	for i, k := range keys {
@@ -102,6 +103,7 @@ func (c *counterImpl) collect() CollectedMetric {
 	for k := range c.values {
 		keys = append(keys, k)
 	}
+	// Sort keys for deterministic output since Go map iteration order is random
 	sort.Strings(keys)
 
 	samples := make([]interface{}, 0, len(c.values))
@@ -188,6 +190,7 @@ func (g *gaugeImpl) collect() CollectedMetric {
 	for k := range g.values {
 		keys = append(keys, k)
 	}
+	// Sort keys for deterministic output since Go map iteration order is random
 	sort.Strings(keys)
 
 	samples := make([]interface{}, 0, len(g.values))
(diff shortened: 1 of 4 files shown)


internal/impactmetrics/metric_types.go
```diff
@@ -1,18 +1,28 @@
...
 )
 
+func isInvalidValue(v float64) bool {
+	return math.IsNaN(v) || math.IsInf(v, 0)
+}
+
 type MetricLabels map[string]string
 
-type IntMetricSample struct {
+type CounterMetricSample struct {
```
kwasniew, line 16: name metrics after their usage and not impl type details

internal/impactmetrics/metric_types.go
```diff
@@ -1,18 +1,28 @@
 package impactmetrics
 
 import (
+	"math"
 	"sort"
 	"strings"
 	"sync"
 )
 
+func isInvalidValue(v float64) bool {
```
kwasniew, line 10: wasn't needed for counter ints but gauge floats introduce new corner cases

internal/impactmetrics/metric_types.go
```diff
@@ -119,14 +131,96 @@ func (c *counterImpl) collect() CollectedMetric {
...
+	defer g.mu.Unlock()
+	key := LabelKey(labels)
+	g.values[key] = value
+}
+
+func (g *gaugeImpl) collect() CollectedMetric {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	keys := make([]string, 0, len(g.values))
```
kwasniew, line 189: we may drop this if we decide sorting in the prod code is not worth it and move sorting to tests instead

internal/impactmetrics/metric_types.go
```diff
@@ -33,6 +43,7 @@ func LabelKey(labels MetricLabels) string {
 	for k := range labels {
 		keys = append(keys, k)
 	}
+	// Sort keys for deterministic output since Go map iteration order is random
```
sighphyre: WDYT?

```suggestion
	// Sort keys for deterministic output since Go map iteration order is random - this is a detail to make testing easier and isn't a requirement for the logic to be correct
```

internal/impactmetrics/metric_types.go
sighphyre: Reasonable comment - this is the kind of thing that helps future readers of the code too and not just PR reviewers

sighphyre (approved):
Seems reasonable

PR #227, opened 2026-01-30
feat: impact metrics histogram

kwasniew:
## About the changes

Adding last metric type histogram. All micro decisions explained inline. The actual change is 250 LOC, the rest is tests.

Also added a new test for all corner case values across all metric types (parametrized test)

### DIFF
--- a/internal/impactmetrics/metric_types.go
+++ b/internal/impactmetrics/metric_types.go
@@ -288,13 +288,10 @@ func (h *histogramImpl) Observe(value float64, labels MetricLabels) {
 	defer h.mu.Unlock()
 
 	key := LabelKey(labels)
-	data, exists := h.values[key]
-	if !exists {
+	data, ok := h.values[key]
+	if !ok {
 		data = &histogramData{
-			buckets: make(map[float64]int64, len(h.buckets)),
-		}
-		for _, b := range h.buckets {
-			data.buckets[b] = 0
+			buckets: make(map[float64]int64),
 		}
 		h.values[key] = data
 	}
@@ -417,7 +414,7 @@ func NewInMemoryMetricRegistry() *InMemoryMetricRegistry {
 func (r *InMemoryMetricRegistry) Counter(name, help string) Counter {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if c, exists := r.counters[name]; exists {
+	if c, ok := r.counters[name]; ok {
 		return c
 	}
 	c := newCounter(name, help)
@@ -428,7 +425,7 @@ func (r *InMemoryMetricRegistry) Counter(name, help string) Counter {
 func (r *InMemoryMetricRegistry) GetCounter(name string) Counter {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	if c, exists := r.counters[name]; exists {
+	if c, ok := r.counters[name]; ok {
 		return c
 	}
 	return nil
@@ -437,7 +434,7 @@ func (r *InMemoryMetricRegistry) GetCounter(name string) Counter {
 func (r *InMemoryMetricRegistry) Gauge(name, help string) Gauge {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if g, exists := r.gauges[name]; exists {
+	if g, ok := r.gauges[name]; ok {
 		return g
 	}
 	g := newGauge(name, help)
@@ -448,7 +445,7 @@ func (r *InMemoryMetricRegistry) Gauge(name, help string) Gauge {
 func (r *InMemoryMetricRegistry) GetGauge(name string) Gauge {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	if g, exists := r.gauges[name]; exists {
+	if g, ok := r.gauges[name]; ok {
 		return g
 	}
 	return nil
@@ -457,7 +454,7 @@ func (r *InMemoryMetricRegistry) GetGauge(name string) Gauge {
 func (r *InMemoryMetricRegistry) Histogram(name, help string, buckets []float64) Histogram {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if h, exists := r.histograms[name]; exists {
+	if h, ok := r.histograms[name]; ok {
 		return h
 	}
 	h := newHistogram(name, help, buckets)
@@ -468,7 +465,7 @@ func (r *InMemoryMetricRegistry) Histogram(name, help string, buckets []float64)
 func (r *InMemoryMetricRegistry) GetHistogram(name string) Histogram {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	if h, exists := r.histograms[name]; exists {
+	if h, ok := r.histograms[name]; ok {
 		return h
 	}
 	return nil
--- a/internal/impactmetrics/metric_types.go
+++ b/internal/impactmetrics/metric_types.go
@@ -548,17 +548,20 @@ func (r *InMemoryMetricRegistry) Restore(metrics []CollectedMetric) {
 			}
 		case "histogram":
 			var buckets []float64
-			if len(m.Samples) > 0 {
-				if first, ok := m.Samples[0].(HistogramMetricSample); ok {
-					for _, b := range first.Buckets {
+			for _, s := range m.Samples {
+				if sample, ok := s.(HistogramMetricSample); ok {
+					for _, b := range sample.Buckets {
 						buckets = append(buckets, bucketLeToFloat(b.Le))
 					}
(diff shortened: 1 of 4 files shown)


internal/impactmetrics/metric_types.go
```diff
@@ -23,13 +23,27 @@ type GaugeMetricSample struct {
 	Value  float64      `json:"value"`
 }
 
+type BucketEntry struct {
+	Le    interface{} `json:"le"`
```
kwasniew: can be "+Inf" or float. Interface to avoid writing custom marshalling

internal/impactmetrics/metric_types.go
```diff
@@ -211,16 +225,192 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+func uniqueSorted(buckets []float64) []float64 {
+	var filtered []float64
+	for _, b := range buckets {
+		if b != math.Inf(1) {
+			filtered = append(filtered, b)
+		}
+	}
+	sort.Float64s(filtered)
+
+	// Deduplicate adjacent elements
```
kwasniew, line 279: I decided not to use slices.Compact which is 1.21+ only. My code will work on older versions too

internal/impactmetrics/metric_types.go
```diff
@@ -211,16 +225,192 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+
+	return &histogramImpl{
+		name:    name,
+		help:    help,
+		buckets: sortedBuckets,
+		values:  map[string]*histogramData{},
+	}
+}
+
+func uniqueSorted(buckets []float64) []float64 {
```
kwasniew, line 270: algorithm here is: filter out inf, sort, deduplicate adjacent. I found it's most idiomatic in go

internal/impactmetrics/metric_types.go
```diff
@@ -211,23 +225,197 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+		counters:   map[string]*counterImpl{},
+		gauges:     map[string]*gaugeImpl{},
+		histograms: map[string]*histogramImpl{},
 	}
 }
 
 func (r *InMemoryMetricRegistry) Counter(name, help string) Counter {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if c, exists := r.counters[name]; exists {
```
kwasniew, line 230: ok should be more idiomatic

internal/impactmetrics/metric_types.go
```diff
@@ -211,23 +225,202 @@ func (g *gaugeImpl) collect() CollectedMetric {
 	}
 }
 
+type Histogram interface {
+	Observe(value float64, labels MetricLabels)
+	Restore(sample HistogramMetricSample)
```
kwasniew, line 238: pragmatic decision similar to node sdk. it makes restore in InMemoryMetricRegistry much simpler. histogram restoration is way more complex than counter and gauge

internal/impactmetrics/metric_types.go
```diff
@@ -320,6 +547,24 @@ func (r *InMemoryMetricRegistry) Restore(metrics []CollectedMetric) {
...
+					}
+					break
+				}
+			}
+
+			h := r.Histogram(m.Name, m.Help, buckets)
+
+			for _, s := range m.Samples {
+				if sample, ok := s.(HistogramMetricSample); ok {
+					h.Restore(sample)
```
kwasniew, line 552: delegate to histogram restore

internal/impactmetrics/metric_types_test.go
```diff
@@ -192,6 +193,181 @@ func TestRestoreReinsertsCollectedMetricsIntoTheRegistry(t *testing.T) {
...
+	tests := []struct {
+		name  string
+		value float64
+	}{
+		{"Infinity", math.Inf(1)},
+		{"-Infinity", math.Inf(-1)},
+		{"NaN", math.NaN()},
+	}
+
+	for _, tc := range tests {
```
kwasniew, line 331: I found that tc may be idiomatic for test case

internal/impactmetrics/metric_types_test.go
```diff
@@ -192,6 +193,181 @@ func TestRestoreReinsertsCollectedMetricsIntoTheRegistry(t *testing.T) {
...
+		},
+	}, emptyCollect)
+
+	registry.Restore(firstCollect)
+
+	restoredCollect := registry.Collect()
+	assert.Equal(t, firstCollect, restoredCollect)
+}
+
+func TestAllMetricOperationsSilentlyDropInvalidValues(t *testing.T) {
```
kwasniew, line 321: now that we have all metric types we can test dropping extreme values

internal/impactmetrics/metric_types.go
```diff
@@ -211,23 +225,202 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+func bucketLeToFloat(le interface{}) float64 {
+	switch v := le.(type) {
+	case string:
+		if v == "+Inf" {
+			return math.Inf(1)
+		}
+	case float64:
+		return v
+	}
+	// Data is internally managed (from CollectedMetric), so invalid types should not occur
```
kwasniew: explaining why no error is thrown

internal/impactmetrics/metric_types.go
```diff
@@ -211,23 +225,202 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+	mu      sync.Mutex
+	name    string
+	help    string
+	buckets []float64
+	values  map[string]*histogramData
+}
+
+func newHistogram(name, help string, buckets []float64) *histogramImpl {
+	if len(buckets) == 0 {
+		buckets = DefaultHistogramBuckets
```
sighphyre: Pretty sure this is safe in this incarnation but referencing the slice directly like this is a footgun waiting to happen if a future patch introduces mutations. A copy here is cheap and a nice future proofing layer
```suggestion
	    buckets = append([]float64(nil), DefaultHistogramBuckets...)
```

internal/impactmetrics/metric_types.go
```diff
@@ -211,23 +225,202 @@ func (g *gaugeImpl) collect() CollectedMetric {
...
+}
+
+func (h *histogramImpl) Observe(value float64, labels MetricLabels) {
+	if isInvalidValue(value) {
+		return
+	}
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	key := LabelKey(labels)
```
sighphyre, line 297: This is safe to move outside of the mutex scope I think and it's probably often the bulk of the work this function does

internal/impactmetrics/metric_types.go
sighphyre: "This will never happen" cases tend to happen with alarming regularity haha. I don't necessarily have a problem the correctness here but casts make the code harder to reason about and they're generally a smell that suggests a better solution is nearby. Some options, which I think improve the clarity of the data flow:

- The most obvious approach here is just to do this conversion just prior to sending. I think this is probably most correct - we shouldn't be converting things into wire format formats in the heart of the system. I'd really rather we do this honestly. Making this correctly marshallable isn't hard - this is effectively already doing that but on a weird layer with side effects
- Use a proper tagged type rather than an interface here. It's not ideal but at least we keep some type safety

If we're not going to do either of those, it's good to get into the habit of returning errors from functions that are failable - it's super safe in Go because you can't blow out your call stack by accident and you give future peeps the chance to do something about errors without having to rework multiple layers

sighphyre (approved):
I don't think this is dangerous. The changes I've asked for could be done later so tacking on approve on here and leaving it to judgement or future work

internal/impactmetrics/metric_types.go
sighphyre: You don't need to worry about that, we don't support anything below 1.21 anyway. Think this is fine either way though

internal/impactmetrics/metric_types.go
kwasniew: good point. will apply this to other function usages for counters and gauges

internal/impactmetrics/metric_types.go
kwasniew: ah of course. excellent point

internal/impactmetrics/metric_types.go
kwasniew: my fear of custom marshaller and unmarshaller led me to suboptimal domain modelling inside this file. Now that I read your comment I think I agree with having clean types in the core and having more boilerplate at the edges.

kwasniew: I will also improve the typig of Samples []interface{} `json:"samples"` since it suffers from the same problem

PR #228, opened 2026-01-30
feat: impact metrics context

kwasniew:
## About the changes

Extracting env from headers. Will be used in the next PR with Metrics API

### DIFF
--- a/internal/impactmetrics/context.go
+++ b/internal/impactmetrics/context.go
@@ -1,0 +1,46 @@
+package impactmetrics
+
+import (
+	"net/http"
+	"strings"
+)
+
+type StaticContext struct {
+	AppName     string
+	Environment string
+}
+
+// Authorization header format: project:environment.hash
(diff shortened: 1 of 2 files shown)


internal/impactmetrics/context.go
```diff
@@ -0,0 +1,46 @@
...
+
+func BuildImpactMetricContext(customHeaders http.Header, appName, environment string) StaticContext {
+	ctx := StaticContext{
+		AppName:     appName,
+		Environment: environment,
+	}
+
+	if customHeaders != nil {
+		if env, ok := extractEnvironmentFromHeaders(customHeaders); ok {
+			ctx.Environment = env
```
kwasniew, line 41: override env property with header. similar pattern to node sdk. at some point we will drop env property then we can change the signature of the BuildImpactMetricContext

sighphyre (approved):
Yes, sick

PR #229, opened 2026-01-30
feat: impact metrics api

kwasniew:
## About the changes

Wrapping all 3 metrics into Metrics API

Important decisions that differ from Node SDK:
* Labels pre-allocated once at construction, not per-call — avoids repeated map allocation on hot path
* Tests use real registry instead of fakes (unlike Node) — verifies actual outcomes not method calls


internal/impactmetrics/metrics_api.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func NewMetricsAPI(metricRegistry *InMemoryMetricRegistry, context StaticContext, warningsChannel chan error) *MetricsAPI {
+	return &MetricsAPI{
+		metricRegistry: metricRegistry,
+		labels: MetricLabels{
+			"appName":     context.AppName,
+			"environment": context.Environment,
+		},
+		warnings: warningsChannel,
```
kwasniew, line 18: I was checking other files and it looks like we use channels for warnings. Java uses logger and Node uses event emitter.

sighphyre: Yep. It's insane and horrible and I hate it but its what we have and you're not making it worse so 👍

internal/impactmetrics/metrics_api.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func (api *MetricsAPI) sendWarning(err error) {
```
sighphyre, line 22: Not sure this is the pattern we want here. If nothing else, it's not what we do when we consume this channel elsewhere. This yeets errors into the ether if the consumer isn't reading them fast enough. Arguably we should do that for errors and warnings, but until we cut over I'd like the guarantees to be the same

internal/impactmetrics/metrics_api.go
```diff
@@ -0,0 +1,82 @@
...
+	select {
+	case api.warnings <- err:
+	default:
+	}
+}
+
+func (api *MetricsAPI) DefineCounter(name, help string) {
+	if name == "" || help == "" {
+		api.sendWarning(fmt.Errorf("counter name and help cannot be empty: name=%s, help=%s", name, help))
```
sighphyre: Make it the same as node? Probably slightly more honest

```suggestion
		api.sendWarning(fmt.Errorf("counter name or help cannot be empty: name=%s, help=%s", name, help))
```

internal/impactmetrics/metrics_api.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (api *MetricsAPI) DefineHistogram(name, help string, buckets ...float64) {
+	if name == "" || help == "" {
+		api.sendWarning(fmt.Errorf("histogram name and help cannot be empty: name=%s, help=%s", name, help))
+		return
+	}
+	api.metricRegistry.Histogram(name, help, buckets)
+}
+
+func (api *MetricsAPI) IncrementCounter(name string, value ...int64) {
```
sighphyre: This is taste an opinion so if you want to keep the poor man's overloading then I can live with it. But I think just having different methods is the better option, it's kinda hard to know what this function wants from reading the signature - we even left ourselves an out in the spec: https://github.com/Unleash/client-specification/blob/main/impact-metrics/impact-metrics.md#public-api!

sighphyre (commented):
Looks good, lemme double check tomorrow with fresh eyes. Sanity check on the warning channel

internal/impactmetrics/metrics_api.go
kwasniew: After doing some research I think separate methods will be more idiomatic for go devs than poor man's overloading with varargs. I will use the leeway we left for ourselves

internal/impactmetrics/metrics_api.go
kwasniew: I'm reading this as let's be consistently wrong rather than inconsistently right. Aaand I think I agree with that :)

sighphyre (approved):
Looks good!

PR #231, opened 2026-02-05
feat: impact metrics integration

kwasniew:
## About the changes

Integrating impact metrics into client API. This is the equivalent of adding ygg bindings to Ruby or Python. 
Had to make maaany decisions and all of them are documented inline.


internal/impactmetrics/metric_types.go
```diff
@@ -24,8 +25,22 @@ type GaugeMetricSample struct {
 }
 
 type BucketEntry struct {
-	Le    float64 `json:"le"`
-	Count int64   `json:"count"`
+	Le    float64
```
kwasniew, line 28: no marshalling annotations since we have custom MarshalJSON for +Inf and I don't want to mislead anyone that we do it with annotations

internal/impactmetrics/metric_types.go
```diff
@@ -24,8 +25,24 @@ type GaugeMetricSample struct {
...
+	Count int64
+}
+
+func (be BucketEntry) MarshalJSON() ([]byte, error) {
+	var leValue interface{}
```
kwasniew, line 33: me trying to do poor man's union type

client.go
```diff
@@ -90,6 +92,16 @@ type metricsChannels struct {
 	registered chan ClientData
 }
 
+func newImpactMetrics(options impactMetricsOptions, channels impactMetricsChannels) (*impactmetrics.MetricsAPI, *impactmetrics.InMemoryMetricRegistry) {
```
kwasniew, line 95: following established pattern of dependency assembly present e.g. in newMetrics

client.go
```diff
@@ -577,6 +601,10 @@ func (uc *Client) Sent() <-chan MetricsData {
 	return uc.sent
 }
 
+func (uc *Client) ImpactMetrics() *impactmetrics.MetricsAPI {
```
kwasniew, line 604: this getter style seems to be idiomatic here

impact_metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestImpactMetricsSentInPayload(t *testing.T) {
+	payloads := make(chan []byte, 1)
```
kwasniew, line 17: channels seem to be a better fit for client/server communication than mutexes

impact_metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	api := client.ImpactMetrics()
+
+	api.DefineCounter("purchases", "Number of purchases")
+	api.IncrementCounter("purchases")
+
+	api.DefineGauge("active_users", "Active users")
+	api.UpdateGauge("active_users", 42)
+
+	api.DefineHistogram("latency", "Request latency", 0.1, 0.5, 1.0)
```
kwasniew, line 50: thinking about changing it to array/slice instead of vararg. Spec advocates for arrays but our go impl may be more idiomatic in go

impact_metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+			w.WriteHeader(http.StatusOK)
+		}
+	}))
+	defer server.Close()
+
+	client, err := NewClient(
+		WithUrl(server.URL),
+		WithAppName("test-app"),
+		WithEnvironment("test"),
+		WithMetricsInterval(5*time.Millisecond),
```
kwasniew, line 37: since we have channels I wanted some very small value here

impact_metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	payloadBytes := <-payloads
+	var payload map[string]interface{}
+	require.NoError(t, json.Unmarshal(payloadBytes, &payload))
+
+	impactMetrics := payload["impactMetrics"].([]interface{})
+	impactMetricsJSON, err := json.Marshal(impactMetrics)
+	require.NoError(t, err)
+
+	expectedJSON := `[
```
kwasniew, line 61: I experimented with different assertions and this one seems to be the most readable. This is the actual stuff on the wire instead of a type infected data structure

metrics.go
```diff
@@ -238,6 +250,10 @@ func (m *metrics) sendMetrics() {
 			m.add(name, false, tc.No)
 		}
 
+		if len(collectedMetrics) > 0 {
+			m.metricRegistry.Restore(collectedMetrics)
```
kwasniew, line 252: registry is self protecting itself with internal locks so locks don't leak out to this file as they do for bucket

README.md
```diff
@@ -287,6 +287,56 @@ func main() {
...
+
+### Histograms
+
+Histograms measure value distribution (request duration, response size):
+
+```go
+client.ImpactMetrics().DefineHistogram(
+    "request_time_ms",
+    "Time taken to process a request in milliseconds",
+    50, 100, 200, 500, 1000,
```
kwasniew, line 332: I'm not sure if slices would be better here

metrics.go
```diff
@@ -200,7 +206,10 @@ func (m *metrics) sendMetrics() {
 	m.bucketMu.Lock()
 	bucket := m.resetBucket()
 	m.bucketMu.Unlock()
-	if bucket.IsEmpty() {
+
+	collectedMetrics := impactmetrics.CollectedMetrics(m.metricRegistry.Collect())
+
+	if bucket.IsEmpty() && collectedMetrics.IsEmpty() {
```
kwasniew, line 212: added IsEmpty to match existing empty check pattern and level of abstraction

impact_metrics_test.go
sighphyre: Yup, if you're coming from Java land you can think of them as thread safe queues (which is more or less what they are)

impact_metrics_test.go
sighphyre: This is arrays behind the scenes anyway. I think this is nice, I'd just leave it alone

internal/impactmetrics/metric_types.go
sighphyre: It's well contained within this function, I think it's fine. Not going to nerd out with suggestions here 😅

impact_metrics_test.go
sighphyre: +1. Also I'd like the tests to be asserting against the wire format anyway so... +2?

sighphyre (approved):
Yeah I like

PR #233, opened 2026-02-06
fix: always send empty variant buckets if not present

### LINKED ISSUE #232: Metrics validation error: variants property must be object but SDK sends null (v6.0.0)
### Describe the bug

After upgrading from `unleash-go-sdk` v5.0.3 to v6.0.0, we're receiving validation errors when the SDK sends metrics to the Unleash server.

**Error Message**
```
Request validation failed: your request body or params contain invalid data. Refer to the details list for more information

details:[{
  "message": "The /body/bucket/toggles/<ff-name>/variants property must be object. You sent null.",
  "path": "/body/bucket/toggles/<ff-name>/variants"
}]
```

**Environment**
- SDK Version: v6.0.0
- Go Version: 1.25

**Configuration**

```
import ucg "github.com/Unleash/unleash-go-sdk/v6"

opts := []ucg.ConfigOption{
    ucg.WithAppName(<app-name>),
    ucg.WithHttpClient(<custom-http-client>),
    ucg.WithListener(<custom-listner>),
    ucg.WithRefreshInterval(<interval>),
    ucg.WithUrl(<url>),
}

c, err := ucg.NewClient(opts...)
```

1. Upgrade from v5.x to v6.0.0
2. Initialize Unleash client with standard configuration (metrics enabled by default)
3. Use a feature toggle that has no variants defined (e.g., simple boolean toggle)
4. Wait for metrics refresh interval
5. Observe validation error in logs

### Expected behavior

The SDK should send a valid metrics payload where the variants field is an empty object {} when a toggle has no variants, not null.

**Actual Behavior**
The SDK sends null for the variants field in the metrics bucket payload, causing the Unleash server to reject the metrics with a validation error.

### Logs, error output, etc.

```bash

```

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

6.10.1

### Subscription type

Open source

### Hosting type

Self-hosted

_No response_

sighphyre:
Unleash expects empty variants lists not nil for whatever reason. V6 changed the internal behavior here to send nils if not present. This patches that to include empties instead.

Fixes #232 

Related to https://github.com/Unleash/unleash/pull/11296

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -233,8 +233,9 @@ func (m *metrics) buildBucketAndReset(lastCloseTime time.Time) (api.Bucket, bool
 		}
 
 		toggleCounters := api.ToggleCount{
-			Yes: int32(yes),
-			No:  int32(no),
+			Yes:      int32(yes),
+			No:       int32(no),
+			Variants: map[string]int32{},
 		}
 
 		// we can have a little locking, as a treat. Variants are likely a luke warm path at best
--- a/metrics_test.go
+++ b/metrics_test.go
@@ -635,3 +635,27 @@ func TestReinsertingBucketsAlsoRestoresVariants(t *testing.T) {
 
 	assert.EqualValues(t, 1, registered_metric.(*toggleCounters).variants["some-variant"])
 }
+
+func TestMetrics_NotCountingVariantsStillIncludesEmptyVariantBucket(t *testing.T) {
+	metrics_handler := &metrics{
+		metricsChannels: metricsChannels{
+			count: make(chan metric, 1),
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


gastonfournier (approved):
LG! Same as we have in node: https://github.com/Unleash/unleash-node-sdk/blob/ce031d4427dcfa2f37188df9fbc67b896877d5b6/src/metrics.ts#L312-L316

sighphyre:
> LG! Same as we have in node: https://github.com/Unleash/unleash-node-sdk/blob/ce031d4427dcfa2f37188df9fbc67b896877d5b6/src/metrics.ts#L312-L316

Yup, thanks for checking!

PR #236, opened 2026-02-06
feat: Impact metrics v6 migration

kwasniew:
## About the changes

Porting impact metrics from v5. No creative coding here, ported line by line. I expected more differences but there were none


sighphyre (approved):
Nice!

PR #239, opened 2026-02-25
feat: implement basic regex constraint

sighphyre:
Basic regex constraint. This is path is a bit expensive and wants some caching. We don't have the prepass compile pattern that the RustSDK/Yggdasil have so I'm planning on slapping an LRU cache on this and moving on with my life -> next PR

### DIFF
--- a/internal/constraints/check.go
+++ b/internal/constraints/check.go
@@ -62,6 +62,8 @@ func checkConstraint(ctx *context.Context, constraint api.Constraint) (bool, err
 		return operatorSemverLt(ctx, constraint)
 	case api.OperatorSemverGt:
 		return operatorSemverGt(ctx, constraint)
+	case api.OperatorRegex:
+		return operatorRegex(ctx, constraint), nil
 	default:
 		return false, fmt.Errorf("unknown constraint operator: %s", constraint.Operator)
 	}
--- a/internal/constraints/operator_regex.go
+++ b/internal/constraints/operator_regex.go
@@ -1,0 +1,24 @@
+package constraints
+
+import (
+	"regexp"
+
+	"github.com/Unleash/unleash-go-sdk/v6/api"
+	"github.com/Unleash/unleash-go-sdk/v6/context"
+)
+
+func operatorRegex(ctx *context.Context, constraint api.Constraint) bool {
+	contextValue := ctx.Field(constraint.ContextName)
+
+	pattern := constraint.Value
+	if constraint.CaseInsensitive {
+		pattern = "(?i:" + pattern + ")"
+	}
+
+	re, err := regexp.Compile(pattern)
+	if err != nil {
+		return false
+	}
+
+	return re.MatchString(contextValue)
+}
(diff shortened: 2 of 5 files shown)


internal/constraints/operator_regex.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/Unleash/unleash-go-sdk/v6/api"
+	"github.com/Unleash/unleash-go-sdk/v6/context"
+)
+
+func operatorRegex(ctx *context.Context, constraint api.Constraint) (bool, error) {
+	contextValue := ctx.Field(constraint.ContextName)
+
+	pattern := constraint.Value
+	if constraint.CaseInsensitive {
+		pattern = "(?i:" + pattern + ")"
```
sighphyre, line 15: Go has no compile options on Regex so we have to do this like some kind of peasant

internal/constraints/operator_regex_test.go
```diff
@@ -0,0 +1,47 @@
+package constraints
```
sighphyre, line 1: 100% vibe coded. Not sure I care much here, the source of truth is the client spec. These are for convenience

internal/constraints/check.go
```diff
@@ -62,6 +62,8 @@ func checkConstraint(ctx *context.Context, constraint api.Constraint) (bool, err
 		return operatorSemverLt(ctx, constraint)
 	case api.OperatorSemverGt:
 		return operatorSemverGt(ctx, constraint)
+	case api.OperatorRegex:
+		return operatorRegex(ctx, constraint), nil
```
sighphyre, line 66: Intentionally not bubbling errors here - our spec says bad regex short circuits to off. In practice not an issue we can encounter since Unleash blocks there by prevalidating the regex with an engine that upholds the same grammar and constraints

internal/constraints/operator_regex.go
gastonfournier: Question: what if the pattern is correct, and by some reason we break it by wrapping it? 

E.g. can the user define their pattern as `"(?i:xyz)"`? I'm thinking of a case where `regexp.Compile(pattern)` works well, but `regexp.Compile("(?i:" + pattern + ")")` does not.

gastonfournier: Still, it will exit by returning false in line 20, but I'm wondering what'd be the right behavior...

gastonfournier (approved):
LG! I think there might be some edge case where the user can make the case-insensitive regexp not work as expected, but I think it'd be an edge case and worst case we can document and fix later.

internal/constraints/operator_regex.go
sighphyre: Hmm... then we would get very unexpected, hard to debug behavior in the SDK

Good callout, let me see if I can do something else here

sighphyre: Okay I can't think of one. ChatGPT says there isn't one. I think this is safe enough for now

PR #240, opened 2026-02-26
chore: regex caching

sighphyre:
NO OPTIMIZATION WITHOUT BENCHMARKING

So yeah, added a quick and dirty bench to get a sense of the hit here. The LRU cache gives us a 10x buff to speed. Which makes this worth it

There's nicer ways to tackle this problem but not without a substantial architectural overall and that's currently not worth it (and probably won't ever be)


internal/constraints/operator_regex.go
```diff
@@ -2,16 +2,48 @@ package constraints
...
+}
+
+var (
+	regexCache     *lru.Cache[regexKey, *regexp.Regexp]
+	regexCacheOnce sync.Once
+)
+
+func getRegexCache() *lru.Cache[regexKey, *regexp.Regexp] {
+	regexCacheOnce.Do(func() {
+		cache, _ := lru.New[regexKey, *regexp.Regexp](100)
```
sighphyre, line 25: 100 is completely arbitrary here. I have no idea of real life workloads but something is substantially better than nothing here

gastonfournier: I think it's reasonable for most workloads. If we have to change it in the future, we can. But this gives a performance boost until you have more than 10 regular expressions.

gastonfournier: Maybe we should just make it unbounded (don't know if possible), or with something similar to a hash map. Because it's like having the compiled expression in other languages.

internal/constraints/operator_regex.go
```diff
@@ -20,5 +52,7 @@ func operatorRegex(ctx *context.Context, constraint api.Constraint) bool {
 		return false
 	}
 
+	cache.Add(key, re)
```
gastonfournier, line 55: If we have a bad regexp, it will never be cached and will always try to compile (which will continuously fail). This will degrade the performance, but I think it's a tolerable thing. As a user, you should be accountable for writing REGEXPs correctly

gastonfournier (approved):

internal/constraints/operator_regex.go
sighphyre: I avoided that because if your upstream changes on a long lived process you can end up with a very large cache of regexes that you'll never use. Also might not be a problem but I'm more willing to accept "sometimes slower than expected" than "omg where RAM go???"

Think this is safe for now. We can always fine tune later

internal/constraints/operator_regex.go
sighphyre: Yes! However, someone was a cunning ginger and considered this - Unleash itself validates the constraint against the same regex dialect before saving it. So we shouldn't be able to get here with the standard flow. 

Could be a bootstrap issue but I think the risk is low - I agree this is on the user at that point

internal/constraints/operator_regex.go
gastonfournier: Yes, I was thinking exactly the same. Arguably, RAM will be used as much as your flag configuration needs, but agree on the current approach

PR #241, opened 2026-02-26
chore: improve ci test formatting

sighphyre:
The test outputs on CI is so large that it causes GH to break in my browser. It's really difficult to see which test is failing and why. I'd like to reduce this down to only basic output and pipe the results into a tool meant to give nice outputs

### DIFF
--- a/.github/workflows/build-prs.yml
+++ b/.github/workflows/build-prs.yml
@@ -2,6 +2,7 @@ name: Build PRs
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


daveleek (approved):
Nice!

sighphyre:
CI is failing because of a coveralls outage. Going to merge this, since it's safe to do. Other active PRs can wait for coveralls to resolve their downtime

PR #243, opened 2026-02-26
chore: force ci format

sighphyre:
Forces formatting on CI because we're bad at that. Also fixes the formatting in metrics and delta because see point 1


gastonfournier (approved):
Good step forward, I'd love to have it on a pre commit hook, but this is good enough, one step at a time :)

PR #245, opened 2026-03-04
fix: uncurse impact metrics api

sighphyre:
API in it's existing state is a bit cursed. First it doesn't expose the impact metrics API in the top level client - that's actually the correct way to use this SDK for most use cases so we need to do that. Second, struct embedding is a first class feature in Go, we should just use it

Touches public API though so needs a deprecation notice + backwards compat

### DIFF
--- a/unleash.go
+++ b/unleash.go
@@ -79,3 +79,51 @@ func Close() error {
 func WaitForReady() {
 	defaultClient.WaitForReady()
 }
+
+// DefineCounter registers a counter metric for impact metrics reporting.
+func DefineCounter(name, help string) {
+	if defaultClient == nil {
+		return
+	}
+	defaultClient.ImpactMetrics().DefineCounter(name, help)
+}
+
+// IncrementCounter increments a previously defined counter metric.
+func IncrementCounter(name string) {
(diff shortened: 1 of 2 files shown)


unleash.go
```diff
@@ -79,3 +79,51 @@ func Close() error {
...
+	defaultClient.DefineCounter(name, help)
+}
+
+// IncrementCounter increments a previously defined counter metric.
+func IncrementCounter(name string) {
+	if defaultClient == nil {
+		return
+	}
+	defaultClient.IncrementCounter(name)
+}
```
gastonfournier, line 97: Should we also expose IncrementCounterBy? Just checking if this is intentional or an oversight: https://github.com/Unleash/unleash-go-sdk/blob/f2c9b51f1ab613c965738282b6d1357394712303/internal/impactmetrics/metrics_api.go#L54

sighphyre: Yes let's do that. I'll patch quickly

gastonfournier (approved):
LGTM!

unleash.go
sighphyre: Done!

PR #201, opened 2025-08-27
Feat/streaming

FredrikOseberg:
Summary

This PR introduces experimental streaming mode support to the
Unleash Go SDK, enabling real-time feature flag updates through
Server-Sent Events (SSE) instead of traditional polling. The
implementation provides instant updates when feature flags change
 on the server.

  Key Features

- Real-time updates: SSE-based streaming eliminates polling delay
 for instant feature flag changes
- Automatic fallback: Seamlessly falls back to polling mode if
streaming is unavailable
- Delta processing: Efficient processing of incremental changes
using hydration and delta events
- Backward compatibility: Existing polling-based clients continue
 to work without changes
- Robust error handling: Proper reconnection logic and error
recovery

# Architecture Changes

##  New Components

- Streaming Client (streaming_client.go): Handles SSE connections
 and event processing
- Streaming Processor (streaming_processor.go): Processes delta
events and manages feature state
- Delta API (api/delta.go): Defines event types for streaming
updates including:
  - HydrationEvent: Full state initialization on connection
  - FeatureUpdatedEvent: Individual feature changes
  - FeatureRemovedEvent: Feature deletions
  - SegmentUpdatedEvent: Segment modifications
  - SegmentRemovedEvent: Segment deletions

## Key Implementation Details

- Uses LaunchDarkly's eventsource library for reliable SSE
connections
- Implements proper delta processing with hydration for initial
state
- Maintains separate streaming and polling code paths for clean
separation
- Clean architecture: Streaming client uses only delta
processing, eliminating redundant full-response logic

## Configuration

Enable streaming mode by setting the experimental streaming
configuration:

unleash.Initialize(
    unleash.WithUrl("https://your-unleash-instance.com/api/"),
    unleash.WithAppName("my-app"),
    unleash.WithExperimental(map[string]interface{}{
        "type": "streaming",
    }),
)

## Testing

- Comprehensive unit tests for streaming components
- Integration tests with mock SSE server
- Delta processing tests covering all event types
- Backward compatibility tests ensuring polling mode continues to
 work

## Files Changed

- Core streaming implementation: 6 new files (~1,200 lines)
- Delta API support: New event types and parsing logic
- Configuration updates: Experimental streaming mode support
- Documentation: Updated README with streaming mode usage
- Dependencies: Added github.com/launchdarkly/eventsource for SSE
 support

PR #247, opened 2026-03-11
chore: trim ci versions to reality

sighphyre:
This PR #201 has pinned us to 1.23 and above. As of Go 1.21, the version specified in go.mod is no longer a hint, it's a requirement. The way our CI runs at the moment will allow for on the fly toolchain upgrades, so 1.21, 1.22 and 1.23 are all practically running on Go version 1.23


gastonfournier (approved):
LGTM! Would be awesome to have the CI fail if we suddenly have a stronger version requirement, but we can start be being honest in here

PR #248, opened 2026-03-13
feat: add semver gte, semver lte and cidr constraints

sighphyre:
Adds Semver LTE, Semver GTE and CIDR constraints

### DIFF
--- a/internal/constraints/operator_semver.go
+++ b/internal/constraints/operator_semver.go
@@ -18,9 +18,21 @@ func operatorSemverLt(ctx *context.Context, constraint api.Constraint) (bool, er
 	})
 }
 
+func operatorSemverLte(ctx *context.Context, constraint api.Constraint) (bool, error) {
+	return operatorSemver(ctx, constraint, func(context *semver.Version, constraint *semver.Version) bool {
+		return context.LessThan(constraint) || context.Equal(constraint)
+	})
+}
+
 func operatorSemverGt(ctx *context.Context, constraint api.Constraint) (bool, error) {
 	return operatorSemver(ctx, constraint, func(context *semver.Version, constraint *semver.Version) bool {
 		return context.GreaterThan(constraint)
+	})
+}
+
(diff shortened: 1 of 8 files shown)


internal/constraints/operator_semver.go
```diff
@@ -18,12 +18,24 @@ func operatorSemverLt(ctx *context.Context, constraint api.Constraint) (bool, er
 	})
 }
 
+func operatorSemverLte(ctx *context.Context, constraint api.Constraint) (bool, error) {
+	return operatorSemver(ctx, constraint, func(context *semver.Version, constraint *semver.Version) bool {
+		return context.LessThan(constraint) || context.Equal(constraint)
```
gastonfournier: Would it be simpler/more efficient to just:

```suggestion
		return !context.GreaterThan(constraint)
```

internal/constraints/operator_semver.go
```diff
@@ -18,12 +18,24 @@ func operatorSemverLt(ctx *context.Context, constraint api.Constraint) (bool, er
...
+
 func operatorSemverGt(ctx *context.Context, constraint api.Constraint) (bool, error) {
 	return operatorSemver(ctx, constraint, func(context *semver.Version, constraint *semver.Version) bool {
 		return context.GreaterThan(constraint)
 	})
 }
 
+func operatorSemverGte(ctx *context.Context, constraint api.Constraint) (bool, error) {
+	return operatorSemver(ctx, constraint, func(context *semver.Version, constraint *semver.Version) bool {
+		return context.GreaterThan(constraint) || context.Equal(constraint)
```
gastonfournier: Same suggestion
```suggestion
		return !context.LessThan(constraint)
```

gastonfournier (approved):
LG just a minor improvement (I think cause it's one less boolean condition and one less numeric comparisson)

internal/constraints/operator_semver.go
sighphyre: I was erring on the side of dumb, readable code. Don't really have strong feelings here so happy to take this, it's a reasonable suggestion

PR #250, opened 2026-03-26
refactor: break up repository

sighphyre:
This starts the break up process between the repository layer and the streaming layer. This is a necessary refactor to get us to the point where we can do failover correctly in this SDK. This PR is very much incomplete work but is an important milestone in the process so putting it here so reviewers can get bite sized chunks of the thought process.

Currently both of these structs are doing too much and their responsibilities are fuzzy. "Repository" is moving towards "pollingFetcher" and "streamingClient" is moving towards "streamingFetcher". Both of these should conform to a shared interface - effectively stop, start and snapshot. This gives us a way to swap these out for each other without a ton of if statements.

### DIFF
--- a/client.go
+++ b/client.go
@@ -196,6 +196,13 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 	headers.Set("unleash-sdk", fmt.Sprintf("%s:%s", clientName, clientVersion))
 	headers.Set("unleash-connection-id", connectionId)
 
+	storage := uc.options.storage
+	if uc.options.storage == nil {
+		storage = &DefaultStorage{}
+	}
+
+	storage.Init(uc.options.backupPath, uc.options.appName)
+
 	uc.repository = newRepository(
 		repositoryOptions{
 			backupPath:      uc.options.backupPath,
@@ -204,7 +211,7 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 			projectName:     uc.options.projectName,
 			instanceId:      uc.options.instanceId,
 			refreshInterval: uc.options.refreshInterval,
-			storage:         uc.options.storage,
+			storage:         storage,
 			httpClient:      uc.options.httpClient,
 			headers:         headers,
 			isStreaming:     uc.options.IsStreamingMode(),
@@ -441,7 +448,7 @@ func (uc *Client) ImpactMetrics() *impactmetrics.MetricsAPI {
 
 // Close stops the client from syncing data from the server.
 func (uc *Client) Close() error {
-	uc.repository.Close()
+	uc.repository.stop()
 	uc.metrics.Close()
 	if uc.options.listener != nil {
 		// Wait for sync to exit.
@@ -510,7 +517,8 @@ func (uc *Client) WaitForReady() {
 
 // ListFeatures returns all available features toggles.
 func (uc *Client) ListFeatures() []api.Feature {
-	return uc.repository.list()
+	snapshot := uc.repository.snapshot()
+	return snapshot.list()
 }
 
 func handleFallback(fallbackFunc FallbackFunc, fallback *bool, featureName string, ctx *context.Context) api.StrategyResult {
--- a/client.go
+++ b/client.go
@@ -45,7 +45,7 @@ var disabledVariantFeatureEnabled = &api.Variant{
 type Client struct {
 	errorChannels
 	options    configOption
-	repository *repository
+	repository togglerFetcher
 	metrics    *metrics
 	*impactmetrics.MetricsAPI
 	strategies         []strategy.Strategy
--- a/client_test.go
+++ b/client_test.go
@@ -1467,8 +1467,6 @@ func TestConnectionAndIntervalHeadersAndBody(t *testing.T) {
 	assert := assert.New(t)
 	defer gock.OffAll()
 
-	gock.Observe(gock.DumpRequest)
-
 	gock.New(mockerServer).
 		Post("/client/register").
 		MatchHeader("UNLEASH-CONNECTION-ID", `[0-9a-f\-]{36}`).
--- a/delta_processor.go
+++ b/delta_processor.go
@@ -3,27 +3,42 @@ package unleash
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Unleash/unleash-go-sdk/v6/api"
 )
 
 type deltaProcessor struct {
-	repository         *repository // ideally this shouldn't be necessary, but for now we're going going to use it to resolve a snapshot of the feature state
-	mu                 sync.RWMutex
-	repositoryChannels repositoryChannels
-	isReady            bool
+	featureState atomic.Value
+	mu           sync.Mutex
 }
 
-func newDeltaProcessor(repo *repository, channels repositoryChannels) *deltaProcessor {
-	return &deltaProcessor{
-		repository:         repo,
-		repositoryChannels: channels,
-		isReady:            false,
+func newDeltaProcessor(baseState *api.FeatureResponse) *deltaProcessor {
+
+	if baseState == nil {
+		baseState = &api.FeatureResponse{
+			Features: make([]api.Feature, 0),
+			Segments: make([]api.Segment, 0),
+		}
 	}
+
+	featureState := &FeatureMemoryState{
+		Features: baseState.FeatureMap(),
+		Segments: baseState.SegmentsMap(),
+	}
+
+	deltaProcessor := &deltaProcessor{
+		featureState: atomic.Value{},
+	}
+
+	deltaProcessor.featureState.Store(featureState)
+	return deltaProcessor
 }
 
-// process processes a delta update from streaming or API events
-// this needs some love, currently it's too intertwined with the repository
+// It's hard to make this both concurrency safe and atomic for both readers and writes. To build our new state we need
+// to hold a lock over the entire process to prevent internal races setting the final state out of order.
+// However, once we have built the new state we can atomically swap it. This gives us mutex locked writes but
+// lock free reads via snapshot()
 func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
 	if delta == nil {
 		return fmt.Errorf("delta is nil")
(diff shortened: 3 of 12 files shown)


client.go
```diff
@@ -196,6 +196,13 @@ func NewClient(options ...ConfigOption) (*Client, error) {
 	headers.Set("unleash-sdk", fmt.Sprintf("%s:%s", clientName, clientVersion))
 	headers.Set("unleash-connection-id", connectionId)
 
+	storage := uc.options.storage
+	if uc.options.storage == nil {
+		storage = &DefaultStorage{}
+	}
+
+	storage.Init(uc.options.backupPath, uc.options.appName)
```
sighphyre, line 204: Done in the client now. This means that the fetching layer doesn't need to care about initializing the storage or checking for nils, which would be silly because it'd be the same code twice

client.go
```diff
@@ -441,7 +448,7 @@ func (uc *Client) ImpactMetrics() *impactmetrics.MetricsAPI {
 
 // Close stops the client from syncing data from the server.
 func (uc *Client) Close() error {
-	uc.repository.Close()
+	uc.repository.stop()
```
sighphyre, line 451: Moving this towards a single shared interface. Close is effectively shutting down the repository at the moment. Two weird things stand out - it's a public method but isn't exposed publicly and really can't be used outside of the core module and secondly the method signature says it returns an error but it never does (not even sure what would make sense there).

It's getting fixed.

client.go
```diff
@@ -510,7 +517,8 @@ func (uc *Client) WaitForReady() {
 
 // ListFeatures returns all available features toggles.
 func (uc *Client) ListFeatures() []api.Feature {
-	return uc.repository.list()
+	snapshot := uc.repository.snapshot()
+	return snapshot.list()
```
sighphyre, line 521: Just a move. Snapshot knows how to `list` now

delta_processor.go
```diff
@@ -3,27 +3,42 @@ package unleash
 import (
```
sighphyre, line 3: Biggest important changes are in this file. DeltaProcessor is now dumb. It has one responsibility - to turn delta events into a rolling snapshot that callers can use to evaluate flags. It knows nothing of backup and nothing of channels

delta_processor.go
```diff
@@ -3,27 +3,42 @@ package unleash
...
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Unleash/unleash-go-sdk/v6/api"
 )
 
 type deltaProcessor struct {
-	repository         *repository // ideally this shouldn't be necessary, but for now we're going going to use it to resolve a snapshot of the feature state
```
sighphyre, line 11: MASTER HAS GIVEN DOBBY AN ATOMIC POINTER, DOBBY IS A FREE ELF

But seriously, closing over a base state and applying updates to that state means the source of truth is here and not in the repository so we don't need it anymore. Finally we get to break the circular dependency

delta_processor.go
```diff
@@ -77,40 +91,37 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
...
 
-	newFeatureList := make([]api.Feature, 0, len(featMap))
-	for _, feature := range featMap {
-		newFeatureList = append(newFeatureList, feature)
+	dp.featureState.Store(&FeatureMemoryState{
+		Features: featMap,
+		Segments: segMap,
+	})
+
+	return nil
```
sighphyre, line 99: Not happy with the error semantics here right now. I'm not sure this is meaningfully fallible. Bigger piece of work - later

delta_processor.go
```diff
@@ -77,40 +91,37 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
...
+}
+
+func (dp *deltaProcessor) snapshot() *FeatureMemoryState {
+	v := dp.featureState.Load()
+	return v.(*FeatureMemoryState)
+}
+
+func (dp *deltaProcessor) asApiResponse() *api.FeatureResponse {
```
sighphyre, line 107: Not yet implemented in a meaningful way but we need a way to represent the goop that we get back from the API - that goes into the backup file. The polling feature has it by default since that... uh... comes from the API. Delta/streaming has to reconstitute it

delta_processor.go
```diff
@@ -77,40 +91,37 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
...
 	}
 
-	state := &api.FeatureResponse{
-		Features: newFeatureList,
-		Segments: newSegmentList,
-	}
-
-	// explicitly ignore error here, saveState fails if we cannot write to the persistent storage
-	// this no longer means that features are in an invalid state so we can enter a ready state
-	_ = dp.repository.saveState(state)
```
sighphyre, line 100: Delta no longer cares about saving state or definitions of ready - that's someone else's problem now

feature_state.go
```diff
@@ -15,6 +15,20 @@ type FeatureMemoryState struct {
 	Segments map[int][]api.Constraint
 }
 
+func (s *FeatureMemoryState) list() []api.Feature {
```
sighphyre, line 18: Just a move

repository.go
```diff
@@ -21,6 +21,12 @@ var (
 	errNoChange = errors.New("no change")
```
sighphyre, line 21: Nothing clever going on here - this disables streaming so that it compiles

repository_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
```
sighphyre, line 5: The tests here were testing polling fetcher specific backoff. That doesn't make sense in streaming. However these tests were going via the outer SDK client - that no longer necessarily has a poll implementation so I've moved these to test that struct directly

streaming_client.go
```diff
@@ -4,33 +4,43 @@ import (
...
 	instanceId         string
 	httpClient         *http.Client
 	headers            http.Header
 	stream             *eventsource.Stream
+	storage            Storage
 	deltaProcessor     *deltaProcessor
 	ctx                context.Context
 	cancel             context.CancelFunc
-	running            bool
-	runningMutex       sync.RWMutex
```
sighphyre, line 26: I've stripped this state + mutex for now. It's trying to represent whether or not the streaming client is running but it's not doing it correctly in a few subtle ways.

streaming_client.go
```diff
@@ -4,33 +4,43 @@ import (
...
 
-// streamingClient handles the SSE connection for streaming feature updates
 type streamingClient struct {
 	url                string
 	appName            string
 	instanceId         string
 	httpClient         *http.Client
 	headers            http.Header
 	stream             *eventsource.Stream
+	storage            Storage
```
sighphyre, line 35: This now knows about storage directly. This leads us grab state at startup to hydrate the delta processor and (later) dump the results to backup on change

streaming_client.go
```diff
@@ -4,33 +4,43 @@ import (
...
 	deltaProcessor     *deltaProcessor
 	ctx                context.Context
 	cancel             context.CancelFunc
-	running            bool
-	runningMutex       sync.RWMutex
 	repositoryChannels repositoryChannels
 }
 
-func newStreamingClient(options repositoryOptions, repoChannels repositoryChannels, deltaProc *deltaProcessor) *streamingClient {
+func newStreamingClient(options repositoryOptions, repoChannels repositoryChannels) *streamingClient {
```
sighphyre, line 57: deltaProcessor can't be meaningfully created outside now. It's bound to the state of the backup. Not sure it needs to either so I've just removed it from the options for now

streaming_client.go
```diff
@@ -40,17 +50,14 @@ func newStreamingClient(options repositoryOptions, repoChannels repositoryChanne
...
 		cancel:             cancel,
 		repositoryChannels: repoChannels,
-		running:            false,
 	}
-}
 
-func (sc *streamingClient) start(_ Storage) error {
-	sc.runningMutex.Lock()
-	defer sc.runningMutex.Unlock()
+	streamingFetcher.sync()
```
sighphyre, line 87: Right now we're just following the patterns in the repository - spin off a goroutine in the construction. Needs to change in the next few PRs

streaming_client.go
```diff
@@ -85,13 +92,6 @@ func (sc *streamingClient) start(_ Storage) error {
 	sc.stream = stream
 
 	go func() {
-		defer func() {
```
sighphyre, line 88: This can't panic unless event processing panics (it can't) or we're handing this a nil channel - but we should validate that elsewhere not in a recover.

The whole point of this refactor is to enable a rework of the error handling semantics in this SDK so I'm just yeeting it for now

streaming_client.go
```diff
@@ -105,68 +105,43 @@ func (sc *streamingClient) start(_ Storage) error {
...
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling connected event: %w", err))
-		}
-	case "unleash-updated":
-		if err := sc.handleUpdatedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling updated event: %w", err))
+	case "unleash-connected", "unleash-updated":
+		if err := sf.handleDomainEvent(event); err != nil {
+			// need to handle failover here but in a future PR
+			// this absolutely cannot just log. Something has gone
+			// very badly wrong and continuing leads to a corrupted state
```
sighphyre, line 185: Next PR. What goes here is the whole reason for this refactor

streaming_client.go
```diff
@@ -105,68 +105,43 @@ func (sc *streamingClient) start(_ Storage) error {
...
-		if err := sc.handleUpdatedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling updated event: %w", err))
+	case "unleash-connected", "unleash-updated":
+		if err := sc.handleDomainEvent(event); err != nil {
+			// need to handle failover here but in a future PR
+			// this absolutely cannot just log. Something has gone
+			// very badly wrong and continuing leads to a corrupted state
 		}
-	default:
-		sc.repositoryChannels.errorChannels.warn(fmt.Errorf("unknown SSE event type: %s", eventType))
```
sighphyre, line 129: Taken this out. This only catches keep alives and other SSE goop we don't care about.

streaming_client.go
```diff
@@ -105,68 +105,43 @@ func (sc *streamingClient) start(_ Storage) error {
...
-	}
-
 	sc.cancel()
 	if sc.stream != nil {
 		sc.stream.Close()
 	}
-	sc.running = false
-}
-
-func (sc *streamingClient) isRunning() bool {
```
sighphyre, line 168: Not public and only used in tests. It's getting yeeted

client_test.go
```diff
@@ -1467,8 +1467,6 @@ func TestConnectionAndIntervalHeadersAndBody(t *testing.T) {
 	assert := assert.New(t)
 	defer gock.OffAll()
 
-	gock.Observe(gock.DumpRequest)
```
sighphyre, line 1470: This dumps the data for literally every request to terminal on any failure. Makes debugging super hard. Its getting removed here because I needed to do this to understand why the tests were failing when I patched them - this was terrible DX

PR #251, opened 2026-03-30
refactor: rework streaming so that it's an alternative to polling, not mixed in alongside

sighphyre:
This is just a refactor of the streaming code to solidify the responsibilities of some components and clean up some mess. This is a necessary step in adding failover because we need to be able to flip between streaming and polling and that's tremendously hard with the current implementation

Specifically:

- Moves delta processor to be a pure processing struct. It now has a single responsibility - to crunch deltas into a state that's appropriate for backup/checking flags
- Repository loses everything related to streaming, step one in renaming it to pollingFetcher
- Streaming client is now self contained, it mimics the pollingFetcher's interface, step one in renaming it to streamingFetcher
- Fixes a bunch of ugly and subtle bugs in the streaming code
- Cleans up tests + yeets a few tests that make no sense


client.go
```diff
@@ -196,25 +196,41 @@ func NewClient(options ...ConfigOption) (*Client, error) {
...
-			headers:         headers,
-			isStreaming:     uc.options.IsStreamingMode(),
-		},
-		repositoryChannels{
-			errorChannels: errChannels,
-			ready:         uc.ready,
-			update:        uc.update,
-		},
-	)
+	storage := uc.options.storage
```
sighphyre, line 199: Slightly different flow in the client now - we setup and initialize storage so the fetching layer doesn't have to care, then we decide if we want a streaming or polling and create the correct fetcher for that here and once

client.go
```diff
@@ -510,7 +526,8 @@ func (uc *Client) WaitForReady() {
 
 // ListFeatures returns all available features toggles.
 func (uc *Client) ListFeatures() []api.Feature {
-	return uc.repository.list()
+	snapshot := uc.repository.snapshot()
+	return snapshot.list()
```
sighphyre, line 530: Not important - this is just a result of moving some functionality into the snapshot struct so that code can live there and not in both streaming and polling code

client_test.go
```diff
@@ -1467,8 +1467,6 @@ func TestConnectionAndIntervalHeadersAndBody(t *testing.T) {
 	assert := assert.New(t)
 	defer gock.OffAll()
 
-	gock.Observe(gock.DumpRequest)
```
sighphyre, line 1470: Small change not related to this PR but a big change. This horrible little line dumps the results of every single HTTP call in the tests to std out in case of failure. That means trawling through thousands of lines to see what went wrong. Couldn't deal with this nonsense while trying to fix tests so I just fixed it

delta_processor.go
```diff
@@ -3,53 +3,67 @@ package unleash
...
+	"sync/atomic"
 
 	"github.com/Unleash/unleash-go-sdk/v6/api"
 )
 
 type deltaProcessor struct {
-	repository         *repository // ideally this shouldn't be necessary, but for now we're going going to use it to resolve a snapshot of the feature state
-	mu                 sync.RWMutex
-	repositoryChannels repositoryChannels
-	isReady            bool
```
sighphyre, line 14: channels + ready state is nonsense. That's a fetching layer responsibility so it got removes from here

delta_processor.go
```diff
@@ -3,53 +3,67 @@ package unleash
...
-		repositoryChannels: channels,
-		isReady:            false,
+func newDeltaProcessor(baseState *api.FeatureResponse) *deltaProcessor {
+
+	if baseState == nil {
+		baseState = &api.FeatureResponse{
+			Features: make([]api.Feature, 0),
+			Segments: make([]api.Segment, 0),
+		}
 	}
```
sighphyre, line 23: This now requires a base state - typically populated by backup or boostrap, all events are applied onto that base state. If you don't set one, you get an empty state - hydration will overwrite that, just being safe really

delta_processor.go
```diff
@@ -3,53 +3,67 @@ package unleash
...
 }
 
-// process processes a delta update from streaming or API events
-// this needs some love, currently it's too intertwined with the repository
-func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
+// It's hard to make this both concurrency safe and atomic for both readers and writes. To build our new state we need
+// to hold a lock over the entire process to prevent internal races setting the final state out of order.
+// However, once we have built the new state we can atomically swap it. This gives us mutex locked writes but
+// lock free reads via snapshot()
+func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) (*api.FeatureResponse, error) {
```
sighphyre, line 42: Delta now crunches the delta and returns a response that's appropriate for dumping to backup. Can't really do that in another call without leaking a lock into a higher layer or introducing a race. Seems like an okay tradeoff

delta_processor.go
```diff
@@ -77,40 +91,17 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
...
+	dp.featureState.Store(newState)
 
-	state := &api.FeatureResponse{
-		Features: newFeatureList,
-		Segments: newSegmentList,
-	}
-
-	// explicitly ignore error here, saveState fails if we cannot write to the persistent storage
-	// this no longer means that features are in an invalid state so we can enter a ready state
-	_ = dp.repository.saveState(state)
```
sighphyre, line 100: All of this ready + write is nonsense that belongs in a different layer - the fetching code specifically

feature_state.go
```diff
@@ -15,6 +15,39 @@ type FeatureMemoryState struct {
 	Segments map[int][]api.Constraint
 }
```
sighphyre, line 16: Everything in this file is just a move - some functionality got put here so I didn't have to duplicate it

metrics_test.go
```diff
@@ -529,8 +529,6 @@ func TestMetrics_metricsData_includes_new_metadata(t *testing.T) {
 	assert := assert.New(t)
 	defer gock.OffAll()
 
-	gock.Observe(gock.DumpRequest)
```
sighphyre, line 532: Same as earlier, just cleaning up tests

repository.go
```diff
@@ -21,24 +21,27 @@ var (
...
+	closed        chan struct{}
+	ctx           context.Context
+	cancel        func()
+	isReady       bool
+	refreshTicker *time.Ticker
+	errors        float64
+	maxSkips      float64
+	skips         float64
+	featureState  atomic.Value // this should always hold an instance of *FeatureMemoryState
 }
```
sighphyre, line 45: Our constructor just drops streaming specific fields, this is now back to being pure polling code

repository.go
```diff
@@ -61,13 +63,6 @@ func newRepository(options repositoryOptions, channels repositoryChannels) *repo
 		repo.options.httpClient = http.DefaultClient
 	}
 
-	if options.storage == nil {
-		repo.options.storage = &DefaultStorage{}
-	}
-
-	repo.options.storage.Init(options.backupPath, options.appName)
-	repo.deltaProcessor = newDeltaProcessor(repo, channels)
```
sighphyre, line 69: All initialized in the client. Which makes sense. Fetch code has no business tinkering with the backup initialization sequence.

repository.go
```diff
@@ -132,46 +119,20 @@ func (r *repository) fetchAndReportError() {
...
-			r.err(fmt.Errorf("failed to start streaming client: %w", err))
-		}
-	}
-
-	// Use polling if not streaming
-	if !isStreaming {
-		r.fetchAndReportError()
-	}
-
+func (r *repository) start() {
```
sighphyre, line 122: Looks like a big change, it's not - this just yeets streaming here

repository.go
```diff
@@ -268,23 +224,6 @@ func (r *repository) statusIsOK(resp *http.Response) error {
 	return fmt.Errorf("%s %s returned status code %d", resp.Request.Method, resp.Request.URL, s)
 }
 
-func (r *repository) list() []api.Feature {
```
sighphyre, line 271: Now owned by featureState itself

repository.go
```diff
@@ -298,10 +237,9 @@ func (r *repository) snapshot() *FeatureMemoryState {
 	return v.(*FeatureMemoryState)
 }
 
-func (r *repository) Close() error {
+func (r *repository) stop() {
```
sighphyre, line 240: This could previously return an error. Except it could never return an error (???). So now it doesn't lie and just says it doesn't return an error (???)

repository_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
```
sighphyre, line 5: I've had to rework some unrelated tests here, they're actually tests surrounding polling backoff. But because those tests go via the client and not the polling repository, there's now no way to get that state off the client. Which is probably good, that's a terrible idea. So now they just new up the polling repository directly and test that

repository_test.go
```diff
@@ -17,6 +18,18 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+type NoOpStorage struct{}
```
sighphyre, line 21: Moved. These belonged to a test suite that tested nothing. I've deleted that suite and salvaged this type

streaming_client_test.go
```diff
@@ -1,158 +0,0 @@
-package unleash
```
sighphyre, line 1: These tests are bad. They're really just testing the delta processor via the streaming client but to do this the streaming client needs to make some ugly compromises around continuing to live even when it's completely broken

These make no sense. We need some proper tests in a future PR but this shape isn't it

streaming_configuration_test.go
```diff
@@ -4,41 +4,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 )
 
-func TestStreamingModeConfiguration(t *testing.T) {
```
sighphyre, line 10: This doesn't appear to do anything useful. In theory it tests config, but that config is already tested later in this file.

streaming_delta_integration_test.go
```diff
@@ -10,22 +10,7 @@ import (
 
 // TestStreamingDeltaIntegration tests the full delta processing flow
```
sighphyre, line 11: Nothing interesting here. This is effectively unchanged, just that some functions now require less inputs and I've removed one test

streaming_client.go
```diff
@@ -4,33 +4,60 @@ import (
...
+	transient streamErrorKind = iota
+	warning
+	fatal
+	corrupted
+)
+
+type streamError struct {
+	kind streamErrorKind
+	err  error
+}
```
sighphyre, line 26: I've created our own internal error types here. These will serve the core of failover in the next PR. These allow us to send errors internally within this struct so we can make a decision around failing over, bubbling the error or just incrementing an error count

streaming_client.go
```diff
@@ -40,21 +67,20 @@ func newStreamingClient(options repositoryOptions, repoChannels repositoryChanne
 		ctx:                ctx,
 		cancel:             cancel,
 		repositoryChannels: repoChannels,
-		running:            false,
```
sighphyre, line 43: This struct was previously trying to track its own internal running state, but it wasn't doing it correctly, there were 3 or 4 cases where this boolean value and the actual running state would be misaligned.

This is the wrong way to handle this problem anyway, this struct now runs until its context is cancelled. That's it. That does mean you can't restart it, but I think that's okay anyway, failover will mean starting a polling fetcher in its stead

streaming_client.go
```diff
@@ -40,21 +67,20 @@ func newStreamingClient(options repositoryOptions, repoChannels repositoryChanne
...
+		internalErrors:     make(chan streamError, 8),
+		storage:            options.storage,
 	}
-}
 
-func (sc *streamingClient) start(_ Storage) error {
-	sc.runningMutex.Lock()
-	defer sc.runningMutex.Unlock()
+	streamingFetcher.start()
```
sighphyre, line 74: Hate this. Constructors shouldn't start async tasks. Currently this mimics the way repository works. I'll change them both in a future PR

streaming_client.go
```diff
@@ -40,21 +67,20 @@ func newStreamingClient(options repositoryOptions, repoChannels repositoryChanne
...
-	if sc.running {
-		return nil
-	}
+	return streamingFetcher
+}
 
+func (sc *streamingClient) start() {
 	req, err := http.NewRequestWithContext(sc.ctx, "GET", sc.url, nil)
 	if err != nil {
-		return fmt.Errorf("failed to create request: %w", err)
```
sighphyre, line 57: Previous behavior was to report an error and then... just... do nothing. This now does into our internal error stream which will give us failover in the next PR

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
 	req.Header.Add("Unleash-Client-Spec", SEGMENT_CLIENT_SPEC_VERSION)
 	req.Header.Add("User-Agent", sc.appName)
 
+	go sc.runStream(req)
+	go sc.superviseStream()
```
sighphyre, line 98: This now moves a bunch of async work into its own channels. Previously this function would just block before mounting any error handling, meaning if you hit an error before while setting up you just fail silently and can't recover.

This is now one goroutine to read infinitely from the stream and a second to respond to errors that may arise from that routine. Means we have a hook to actually do failover

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-		return fmt.Errorf("failed to establish streaming connection: %w", err)
+		sc.internalErrors <- streamError{kind: fatal, err: fmt.Errorf("failed to subscribe to SSE stream: %w", err)}
+		return
 	}
 
 	sc.stream = stream
 
-	go func() {
-		defer func() {
-			if r := recover(); r != nil {
```
sighphyre, line 89: This panic recovery is meaningless because nothing here can panic

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-				err := fmt.Errorf("SSE subscription panic recovered: %v", r)
-				sc.repositoryChannels.errorChannels.err(err)
+	for {
+		select {
+		case event, ok := <-stream.Events:
+			if !ok {
```
sighphyre, line 122: We were previously doing this:

``` go
	select {
			case event := <-stream.Events:
				if event != nil {
					sc.handleEvent(event)
				}
     }
```

Which is a little bit wild. This doesn't detect a closed stream so it just spins on a single goroutine forever as fast as possible and noms a CPU. If the channel is closed now, we terminate this goroutine

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-
-		for {
-			select {
-			case event := <-stream.Events:
-				if event != nil {
-					sc.handleEvent(event)
+			switch event.Event() {
+			case "unleash-connected", "unleash-updated":
+				if err := sc.handleDomainEvent(event); err != nil {
+					sc.internalErrors <- streamError{kind: corrupted, err: fmt.Errorf("failed to handle event: %w", err)}
```
sighphyre, line 128: If you miss an event, you're done. Previously we'd log an error and continue. But this leads to a corrupted state, which needs to trigger reconnection - that's coming in the next PR

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-	}()
-
-	sc.running = true
-	return nil
+	}
 }
 
-func (sc *streamingClient) handleEvent(event eventsource.Event) {
-	eventType := event.Event()
+func (sc *streamingClient) superviseStream() {
```
sighphyre, line 138: Not implemented. Coming next

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-	case "unleash-connected":
-		if err := sc.handleConnectedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling connected event: %w", err))
-		}
-	case "unleash-updated":
-		if err := sc.handleUpdatedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling updated event: %w", err))
-		}
-	default:
-		sc.repositoryChannels.errorChannels.warn(fmt.Errorf("unknown SSE event type: %s", eventType))
```
sighphyre, line 129: Erroneously warns on keep alive events

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
+func (sc *streamingClient) notifyUpdate() {
+	if sc.hydrated.CompareAndSwap(false, true) {
+		sc.ready <- true
+	} else {
+		sc.update <- true
 	}
 }
 
-func (sc *streamingClient) handleConnectedEvent(event eventsource.Event) error {
+func (sc *streamingClient) handleDomainEvent(event eventsource.Event) error {
```
sighphyre, line 162: I've collapsed these two functions. It's the same code, doesn't need to be a different function

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-	if !sc.running {
-		return
+	// Failure to save a backup is definitely not an error but end users should have
+	// a way of detecting that this isn't working correctly so we throw it on the warnings channel
+	// this just bypasses the internal error handling mechanisms because this never needs to inform
+	// failover or reconnection strategies
+	if err := sc.storage.Persist(backupState); err != nil {
+		select {
+		case sc.warnings <- fmt.Errorf("failed to persist state: %w", err):
+		default:
```
sighphyre, line 184: Can't remember what the warn channel semantics are but I'm not sure it matters - this is a non blocking channel write. If the channel is full instead of blocking this just drops the warnings. I think that's okay

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-		if err := sc.handleConnectedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling connected event: %w", err))
-		}
-	case "unleash-updated":
-		if err := sc.handleUpdatedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling updated event: %w", err))
-		}
-	default:
-		sc.repositoryChannels.errorChannels.warn(fmt.Errorf("unknown SSE event type: %s", eventType))
+func (sc *streamingClient) notifyUpdate() {
```
sighphyre, line 154: Just a lock free ready/update. No need for a mutex here, that's complete overkill. I'll probably patch the polling code later to do something similar

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-	sc.cancel()
-	if sc.stream != nil {
-		sc.stream.Close()
-	}
-	sc.running = false
+	return nil
+}
+
+func (sc *streamingClient) snapshot() *FeatureMemoryState {
+	return sc.deltaProcessor.snapshot()
```
sighphyre, line 192: Similar to polling code - we expose a way to get the snapshot so the client can do isEnabled work. This is an atomic primitive so read access is lock free and fast. Write access is internal to the deltaprocessor and correctly tolerant to data races

delta_processor.go
```diff
@@ -3,53 +3,67 @@ package unleash
...
-// this needs some love, currently it's too intertwined with the repository
-func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) error {
+// It's hard to make this both concurrency safe and atomic for both readers and writes. To build our new state we need
+// to hold a lock over the entire process to prevent internal races setting the final state out of order.
+// However, once we have built the new state we can atomically swap it. This gives us mutex locked writes but
+// lock free reads via snapshot()
+func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) (*api.FeatureResponse, error) {
 	if delta == nil {
-		return fmt.Errorf("delta is nil")
+		return nil, fmt.Errorf("delta is nil")
```
gastonfournier, line 44: If I understand this correctly, it is now properly returning a nil response with an error instead of just returning the error message as a successful response.

streaming_client.go
```diff
@@ -4,33 +4,60 @@ import (
...
 
 	"github.com/Unleash/unleash-go-sdk/v6/api"
 	"github.com/launchdarkly/eventsource"
 )
 
-// streamingClient handles the SSE connection for streaming feature updates
+type streamErrorKind int
+
+const (
+	transient streamErrorKind = iota
```
gastonfournier, line 17: TIL about iota, Golang is weird :D

streaming_client.go
```diff
@@ -68,105 +94,104 @@ func (sc *streamingClient) start(_ Storage) error {
...
-		if err := sc.handleUpdatedEvent(event); err != nil {
-			sc.repositoryChannels.errorChannels.err(fmt.Errorf("error handling updated event: %w", err))
-		}
-	default:
-		sc.repositoryChannels.errorChannels.warn(fmt.Errorf("unknown SSE event type: %s", eventType))
+func (sc *streamingClient) notifyUpdate() {
+	if sc.hydrated.CompareAndSwap(false, true) {
+		sc.ready <- true
+	} else {
+		sc.update <- true
```
gastonfournier, line 158: Should we make both sends non-blocking?

Sending to a channel blocks when the channel buffer is full. Both channels are buffered with capacity 1, so if the consumer has not processed the previous notification yet, a plain `sc.ready <- true` or `sc.update <- true` can pause the goroutine that is handling streaming events. That means a burst of SSE messages, or a slow listener, could stall the stream-processing path.

```
  func (sc *streamingClient) notifyUpdate() {
  	if sc.hydrated.CompareAndSwap(false, true) {
  		select {
  		case sc.ready <- true:
  		default:
  		}
  		return
  	}

  	select {
  	case sc.update <- true:
  	default:
  	}
  }
```

gastonfournier (approved):
LG, just one small comment that might or might not be relevant, I'm leaving that up to you, who have more context. Worst case, it adds some waiting time.

delta_processor.go
sighphyre: Not quite here, this is a change that I've made to allow for concurrency safe file updates. Previously the update here and the file write weren't controlled by the same concurrency primitives, meaning if you were unlucky, what was stored in the backup could be out of date. This change here means we're now returning a data blob appropriate for backup file write within the same bounds as the mutex that updated it

The thing you're thinking about is here; https://github.com/Unleash/unleash-go-sdk/pull/252#discussion_r3013986884 :sweat_smile:

streaming_client.go
sighphyre: So weird

streaming_client.go
sighphyre: This is a good spot but it makes the mistake of assuming that our SDK is sane, haha. This follows the same patterns in the polling code, which in theory can block. We solve that problem by enforcing that there's always a reader that drains these channels

This isn't a sane pattern but you work with the architecture you have and not the architecture you want :sob:

PR #252, opened 2026-03-30
refactor: repository start and construction is now split

sighphyre:
Just a bit of hygiene. Construction a repository no longer automagically starts the fetching goroutine, you now need to call start. Also cleans up some polling behavior where we were using errors to signal a changed event


repository.go
```diff
@@ -156,12 +145,12 @@ func (r *repository) configurationError() {
 	r.skips = r.errors
 }
 
-func (r *repository) fetch() error {
+func (r *repository) fetch() (bool, error) {
```
sighphyre, line 148: Just some clean up because this code is definitely special. This was returning values in the error, now it just returns values in the value

gastonfournier (approved):
LGTM!

PR #253, opened 2026-03-31
refactor: rename fetchers

sighphyre:
This is just a big sweeping change to rename repository + streamingClient to pollingFetcher + streamingFetcher respectively. They now do what they say on the tin and no more. 

This doesn't rename RepositoryOptions because those are publically exposed and need a major version bump to patch


gastonfournier (approved):
LGTM

PR #254, opened 2026-04-01
feat: introduce adaptiveFetcher so we can include failover in streaming

sighphyre:
Basic adaptive fetcher implementation that should allow us to do streaming failover. Not wired up, that's done in #255.

Go's per-emptive concurrency model makes synchronizing this correctly with stop/start a huge pain in the butt so I've chosen to keep this internal only - you make an adaptiveFetcher and it does it's thing, you're not intended to pass in polling+streaming fetchers.This means fetchers continue to be start once, stop once, never restart

Currently the adaptiveFetcher doesn't wait for readiness signals from streaming before cutting over to the pollingFetcher. I'm aware, it's coming in the next set of PRs. With most setups this doesn't matter since it'll read from backup. Will be fixed before release

### DIFF
--- a/adaptive_fetcher.go
+++ b/adaptive_fetcher.go
@@ -1,0 +1,151 @@
+package unleash
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// This exists purely because atomics must contain a single concrete type
+// we want to be able to both switch fetcher types and avoid mutexes so this
+// leaves us with this sad little container struct. Don't "fix" this
+type fetchContainer struct {
+	f togglerFetcher
+}
+
+type fetcherFactory struct {
+	newStreaming func(fetcherOptions, fetcherChannels) togglerFetcher
+	newPolling   func(fetcherOptions, fetcherChannels) togglerFetcher
+}
+
+func defaultFetcherFactory() fetcherFactory {
+	return fetcherFactory{
+		newStreaming: func(options fetcherOptions, channels fetcherChannels) togglerFetcher {
+			return newStreamingFetcher(options, channels)
+		},
+		newPolling: func(options fetcherOptions, channels fetcherChannels) togglerFetcher {
+			return newPollingFetcher(options, channels)
+		},
+	}
+}
+
+type adaptiveFetcher struct {
+	currentFetcher atomic.Pointer[fetchContainer]
+	baseOptions    fetcherOptions
+	baseChannels   fetcherChannels
+	internalReady  chan bool
+	internalUpdate chan bool
+	ready          atomic.Bool
+	started        bool
+	stopped        bool
+	ctx            context.Context
+	cancel         context.CancelFunc
+	mu             sync.Mutex
+	factory        fetcherFactory
+}
+
+func newAdaptiveFetcher(options fetcherOptions, channels fetcherChannels) *adaptiveFetcher {
+	return newAdaptiveFetcherWithFactory(options, channels, defaultFetcherFactory())
+}
+
+// If you're considering using this in production code, probably don't. This is exposed
+// internally so that test code has a seam to inject fake fetchers and control the lifecycle.
+// Prefer using newAdaptiveFetcher, which will construct appropriate fetchers for
+// real world usage
+func newAdaptiveFetcherWithFactory(
+	options fetcherOptions,
+	channels fetcherChannels,
+	factory fetcherFactory,
+) *adaptiveFetcher {
+	ir := make(chan bool, 1)
+	iu := make(chan bool, 1)
+
+	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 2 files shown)


adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+type fetchContainer struct {
+	f togglerFetcher
+}
+
+type fetcherFactory struct {
+	newStreaming func(fetcherOptions, fetcherChannels) togglerFetcher
+	newPolling   func(fetcherOptions, fetcherChannels) togglerFetcher
+}
+
+func defaultFetcherFactory() fetcherFactory {
```
sighphyre, line 21: Don't love this but making a seam here is better than not testing so meh

adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+// internally so that test code has a seam to inject fake fetchers and control the lifecycle.
+// Prefer using newAdaptiveFetcher, which will construct appropriate fetchers for
+// real world usage
+func newAdaptiveFetcherWithFactory(
+	options fetcherOptions,
+	channels fetcherChannels,
+	factory fetcherFactory,
+) *adaptiveFetcher {
+	ir := make(chan bool, 1)
+	iu := make(chan bool, 1)
```
sighphyre, line 61: To manage update and readiness we pass our own channels into the downstream fetchers. This means they signal to these channels, which we read and then we make our own decision about what to with those signals

adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return af
+}
+
+func (af *adaptiveFetcher) superviseFetchers() {
+	for {
+		select {
+		case <-af.internalReady:
+			if af.ready.CompareAndSwap(false, true) {
+				af.baseChannels.ready <- true
```
sighphyre, line 91: If we're ready then we're ready. Swapping to a new fetcher means a new readiness signal, which we dedupe here

adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+			return newStreamingFetcher(options, channels)
+		},
+		newPolling: func(options fetcherOptions, channels fetcherChannels) togglerFetcher {
+			return newPollingFetcher(options, channels)
+		},
+	}
+}
+
+type adaptiveFetcher struct {
+	currentFetcher atomic.Pointer[fetchContainer]
```
sighphyre, line 33: We store our fetcher in an atomic pointer. This means we get access to feature flags lock free. But it doesn't prevent us from race conditions in the swap/start/stop so we still gotta mutex this in the write paths

adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+	next := af.factory.newPolling(af.baseOptions, fetcherChannels{
+		ready:         af.internalReady,
+		update:        af.internalUpdate,
+		errorChannels: af.baseChannels.errorChannels,
+	})
+
+	af.currentFetcher.Store(&fetchContainer{f: next})
+
+	next.start()
+	current.stop()
```
daveleek, line 123: Flagging a potential race here, esp if we use delta with polling in the future

adaptive_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+		return
+	}
+
+	next := af.factory.newPolling(af.baseOptions, fetcherChannels{
+		ready:         af.internalReady,
+		update:        af.internalUpdate,
+		errorChannels: af.baseChannels.errorChannels,
+	})
+
+	af.currentFetcher.Store(&fetchContainer{f: next})
```
gastonfournier, line 120: Do we have to delay this until the next fetcher is ready? I'm thinking:
1. You evaluate a flag -> evaluates to true
2. You failover
3. You evaluate a flag -> defaults to disabled
4. Polling gets ready
5. You evaluate a flag -> evaluates to true

Not sure whether this is possible or not, it seems like it

sighphyre: Yessir. Pointed out in the PR description haha

> Currently the adaptiveFetcher doesn't wait for readiness signals from streaming before cutting over to the pollingFetcher. I'm aware, it's coming in the next set of PRs. With most setups this doesn't matter since it'll read from backup. Will be fixed before release

gastonfournier (approved):
LG with one small comment but we discussed offline we have a plan for this, which will come later

adaptive_fetcher.go
sighphyre: There's a race in that it might not be ready when we pass control over, and that's only because fetchers start in a separate Goroutine. Aware of the issue and pointed out in the PR description! It's a smaller problem than it seems though because in practice, by the time you've created the polling fetcher it's hydrated from the backup and is ready to go

PR #255, opened 2026-04-01
chore: wiring adaptive fetching to streaming

sighphyre:
Wire adaptive fetching and streaming fetcher together. We do this by passing a channel for the streaming fetcher to send error signals that are specifically related to streaming

### DIFF
--- a/adaptive_fetcher.go
+++ b/adaptive_fetcher.go
@@ -14,14 +14,14 @@ type fetchContainer struct {
 }
 
 type fetcherFactory struct {
-	newStreaming func(fetcherOptions, fetcherChannels) togglerFetcher
+	newStreaming func(fetcherOptions, fetcherChannels, chan streamError) togglerFetcher
 	newPolling   func(fetcherOptions, fetcherChannels) togglerFetcher
 }
 
 func defaultFetcherFactory() fetcherFactory {
 	return fetcherFactory{
-		newStreaming: func(options fetcherOptions, channels fetcherChannels) togglerFetcher {
-			return newStreamingFetcher(options, channels)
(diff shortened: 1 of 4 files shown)


adaptive_fetcher.go
```diff
@@ -30,18 +30,19 @@ func defaultFetcherFactory() fetcherFactory {
...
-	ctx            context.Context
-	cancel         context.CancelFunc
-	mu             sync.Mutex
-	factory        fetcherFactory
+	currentFetcher        atomic.Pointer[fetchContainer]
+	baseOptions           fetcherOptions
+	baseChannels          fetcherChannels
+	internalReady         chan bool
+	internalUpdate        chan bool
+	failoverSignalChannel chan streamError
```
sighphyre, line 38: Internal update is now failoverSignalChannel. This allows us to tell somebody when something went wrong if they want to subscribe to that channel

adaptive_fetcher.go
```diff
@@ -92,6 +95,8 @@ func (af *adaptiveFetcher) superviseFetchers() {
 			}
 		case <-af.internalUpdate:
 			af.baseChannels.update <- true
+		case <-af.failoverSignalChannel:
+			// do nothing for now but this is where cutover will get triggered from
```
sighphyre, line 99: Moving the blackhole here but still a blackhole

streaming_fetcher.go
```diff
@@ -133,22 +129,6 @@ func (sc *streamingFetcher) runStream(req *http.Request) {
 	}
 }
 
-func (sc *streamingFetcher) superviseStream() {
```
sighphyre, line 136: Don't care anymore. Someone else's job to care now. Typically the adaptive fetcher but whoever wants to care can care

gastonfournier (approved):
LGTM!

PR #256, opened 2026-04-01
feat: add in failover strategy

sighphyre:
More or less a direct port of the Node/Java/C# SDKs' failover strategy. Main code is human rolled and a little adapted to be more Go idiomatic. Test code was done by pointing Codex at the Node tests and asking for a port - human verified but AI rolled


gastonfournier (approved):
LGTM!

PR #257, opened 2026-04-02
feat: wire up adaptive and streaming fetchers

sighphyre:
Wires up the streaming and adaptive fetchers. This means streaming can now signal to the adaptive fetcher that it's out and wants to fail over.

### DIFF
--- a/adaptive_fetcher.go
+++ b/adaptive_fetcher.go
@@ -2,6 +2,7 @@ package unleash
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 )
@@ -14,14 +15,14 @@ type fetchContainer struct {
 }
 
 type fetcherFactory struct {
-	newStreaming func(fetcherOptions, fetcherChannels, chan streamError) togglerFetcher
+	newStreaming func(fetcherOptions, fetcherChannels, chan failEvent) togglerFetcher
 	newPolling   func(fetcherOptions, fetcherChannels) togglerFetcher
 }
 
 func defaultFetcherFactory() fetcherFactory {
 	return fetcherFactory{
-		newStreaming: func(options fetcherOptions, channels fetcherChannels, streamErrorChannel chan streamError) togglerFetcher {
-			return newStreamingFetcher(options, channels, streamErrorChannel)
+		newStreaming: func(options fetcherOptions, channels fetcherChannels, failoverChannel chan failEvent) togglerFetcher {
+			return newStreamingFetcher(options, channels, failoverChannel)
 		},
 		newPolling: func(options fetcherOptions, channels fetcherChannels) togglerFetcher {
 			return newPollingFetcher(options, channels)
@@ -35,7 +36,7 @@ type adaptiveFetcher struct {
 	baseChannels          fetcherChannels
 	internalReady         chan bool
 	internalUpdate        chan bool
-	failoverSignalChannel chan streamError
+	failoverSignalChannel chan failEvent
 	ready                 atomic.Bool
 	started               bool
 	stopped               bool
@@ -60,7 +61,7 @@ func newAdaptiveFetcherWithFactory(
 ) *adaptiveFetcher {
 	ir := make(chan bool, 1)
 	iu := make(chan bool, 1)
-	se := make(chan streamError, 1)
+	fc := make(chan failEvent, 1)
 
 	ctx, cancel := context.WithCancel(context.Background())
 
@@ -72,7 +73,7 @@ func newAdaptiveFetcherWithFactory(
 		ctx:                   ctx,
 		cancel:                cancel,
 		factory:               factory,
-		failoverSignalChannel: se,
+		failoverSignalChannel: fc,
 	}
 
 	initial := af.factory.newStreaming(options, fetcherChannels{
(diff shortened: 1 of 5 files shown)


adaptive_fetcher.go
```diff
@@ -14,14 +15,14 @@ type fetchContainer struct {
 }
 
 type fetcherFactory struct {
-	newStreaming func(fetcherOptions, fetcherChannels, chan streamError) togglerFetcher
+	newStreaming func(fetcherOptions, fetcherChannels, chan failEvent) togglerFetcher
```
sighphyre, line 18: Just unifying the error types. No reason to have two, was a bit of overdesign in my previous PRs

adaptive_fetcher_test.go
```diff
@@ -183,3 +202,51 @@ func TestAdaptiveFetcher_SnapshotAfterCutover(t *testing.T) {
...
+	}
+
+	waitFor(t, 500*time.Millisecond, func() bool {
+		return polling.startCalls == 1 && streaming.stopCalls == 1
+	}, "expected cutover to polling after failover signal")
+
+	select {
+	case err := <-channels.warnings:
+		if err == nil || err.Error() == "" {
+			t.Fatalf("expected warning error on failover")
```
sighphyre, line 248: I want to know a warning was emitted - that's important I think. But I don't want to concretize the actual message. Annoys me when I clean up strings and tests fails

failover.go
```diff
@@ -27,9 +27,8 @@ type httpStatusErrorEvent struct {
 	statusCode int
 }
 
-type serverEvent struct {
+type failoverRequest struct {
```
sighphyre, line 30: Given this one a rename. Previously this was just for the server event requesting a failover, but that's silly, this is now for anything that wants to shortcircuit the failover process and yeet to polling. Example, broken config

failover_test.go
```diff
@@ -10,36 +10,18 @@ func TestFailoverStrategy_ServerHintPollingFailsOver(t *testing.T) {
...
 		},
-		event: "polling",
 	}
 
 	if got := strategy.shouldFailover(event, now); !got {
 		t.Fatalf("expected failover on server polling hint")
 	}
 }
 
-func TestFailoverStrategy_ArbitraryServerEventDoesNotFailover(t *testing.T) {
```
sighphyre, line 26: No longer makes sense

streaming_fetcher.go
```diff
@@ -4,43 +4,45 @@ import (
...
+	headers         http.Header
+	storage         Storage
+	deltaProcessor  *deltaProcessor
+	ctx             context.Context
+	cancel          context.CancelFunc
+	hydrated        atomic.Bool
+	recordFailEvent func(failEvent)
+}
+
+func buildFailoverRecorder(ch chan failEvent, failoverStrategy *failoverStrategy) func(failEvent) {
```
sighphyre, line 30: Don't really want to hold disparate and complex state in the shape of the failover strategy + channel + once lock for that. So it gets wrapped in a closure. Yes it could be a struct. No I don't really care. Closure feels reasonably nice here

streaming_fetcher.go
```diff
@@ -4,43 +4,45 @@ import (
...
+	if ch == nil {
+		return func(failEvent) {}
+	}
+
+	var once sync.Once
+	return func(reason failEvent) {
+		if failoverStrategy.shouldFailover(reason, time.Now()) {
+			once.Do(func() {
+				ch <- reason
```
sighphyre, line 39: I want to tell the caller why we're yeeting streaming so I need a message. Typical Go pattern would be a one shot channel that gets closed but that pattern doesn't work with values like this

streaming_fetcher.go
```diff
@@ -57,17 +59,17 @@ func newStreamingFetcher(options fetcherOptions, fetcherChannels fetcherChannels
...
+		url:             fmt.Sprintf("%sclient/streaming", options.url.String()),
+		appName:         options.appName,
+		instanceId:      options.instanceId,
+		httpClient:      options.httpClient,
+		headers:         options.headers,
+		deltaProcessor:  deltaProc,
+		ctx:             ctx,
+		cancel:          cancel,
+		fetcherChannels: fetcherChannels,
+		recordFailEvent: buildFailoverRecorder(failoverChannel, newFailoverStrategy(5, 60*time.Second)),
```
sighphyre, line 71: Same defaults as Node. Not sure I want to move this to config in the early days because that means altering a public interface. Think this is okay until we're happy this is stable and we see a reason to change it

streaming_fetcher.go
```diff
@@ -100,13 +102,37 @@ func (sc *streamingFetcher) runStream(req *http.Request) {
 		eventsource.StreamOptionUseBackoff(5*time.Minute),
 		eventsource.StreamOptionUseJitter(0.5),
 		eventsource.StreamOptionErrorHandler(func(err error) eventsource.StreamErrorHandlerResult {
-			sc.signalStreamingError(streamError{kind: transient, err: fmt.Errorf("SSE error: %w", err)})
+
+			var event failEvent = nil
+
+			if se, ok := err.(eventsource.SubscriptionError); ok {
```
sighphyre, line 108: Thank you, LD, nice docs on how to disambiguate the error type and nice design around getting the inner values out to get an HTTP failure

gastonfournier (approved):
Seems reasonable to me

PR #259, opened 2026-04-02
chore: small cleanups

sighphyre:
Tooling got updated, it's asking for a few small patches

No idea when codex started generating .codex files but apparently that's a thing now too so it's going in the .gitignore

### DIFF
--- a/feature_state.go
+++ b/feature_state.go
@@ -105,7 +105,7 @@ func (s *FeatureMemoryState) evaluateFeature(
 			continue
 		}
 
-		if stCfg.Variants != nil && len(stCfg.Variants) > 0 {
+		if len(stCfg.Variants) > 0 {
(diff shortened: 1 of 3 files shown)


feature_state.go
```diff
@@ -105,7 +105,7 @@ func (s *FeatureMemoryState) evaluateFeature(
 			continue
 		}
 
-		if stCfg.Variants != nil && len(stCfg.Variants) > 0 {
+		if len(stCfg.Variants) > 0 {
```
sighphyre, line 108: Len is safe on nils

gastonfournier (approved):
This seems to be repeated, I already reviewed this code in a larger PR :D

sighphyre:
> This seems to be repeated, I already reviewed this code in a larger PR :D

I forgot to change the base! Yeah I want to do them in separate steps so that this is easy to roll back. Just in case

I'll switch the base on the other before merge

PR #262, opened 2026-04-02
feat: restart poisoned streams instead of leaving corrupted state

sighphyre:
This terminates a stream and restarts it in the case of poisoned events. In theory this should never happen, however, if for some reason the SDK receives a domain event it cannot process then continuing on is not safe - a rehydration from scratch is necessary to maintain data integrity

### DIFF
--- a/streaming_fetcher.go
+++ b/streaming_fetcher.go
@@ -14,17 +14,24 @@ import (
 
 type streamingFetcher struct {
 	fetcherChannels
-	url             string
-	appName         string
-	instanceId      string
-	httpClient      *http.Client
-	headers         http.Header
-	storage         Storage
-	deltaProcessor  *deltaProcessor
-	ctx             context.Context
-	cancel          context.CancelFunc
+	url            string
+	appName        string
+	instanceId     string
+	httpClient     *http.Client
+	headers        http.Header
+	storage        Storage
+	deltaProcessor *deltaProcessor
+
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	streamMu     sync.Mutex
+	streamCancel context.CancelFunc
+
 	hydrated        atomic.Bool
 	recordFailEvent func(failEvent)
+	subscribe       func(*http.Request, ...eventsource.StreamOption) (*eventsource.Stream, error)
+	restartStreamFn func()
 }
 
 func buildFailoverRecorder(ch chan failEvent, failoverStrategy *failoverStrategy) func(failEvent) {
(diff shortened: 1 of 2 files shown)


streaming_fetcher.go
```diff
@@ -14,17 +14,24 @@ import (
...
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	streamMu     sync.Mutex
+	streamCancel context.CancelFunc
+
 	hydrated        atomic.Bool
 	recordFailEvent func(failEvent)
+	subscribe       func(*http.Request, ...eventsource.StreamOption) (*eventsource.Stream, error)
+	restartStreamFn func()
```
sighphyre, line 34: Don't see a way to test this behavior nicely without introducing this seam. Not a huge fan but I also don't think it's a disaster

streaming_fetcher.go
```diff
@@ -14,17 +14,24 @@ import (
...
+	httpClient     *http.Client
+	headers        http.Header
+	storage        Storage
+	deltaProcessor *deltaProcessor
+
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	streamMu     sync.Mutex
+	streamCancel context.CancelFunc
```
sighphyre, line 29: We now have stream level cancellation, which allows us to yeet the stream and start a new one without breaking the struct's guarantees

streaming_fetcher.go
```diff
@@ -145,11 +177,10 @@ func (sc *streamingFetcher) runStream(req *http.Request) {
 			switch event.Event() {
 			case "unleash-connected", "unleash-updated":
 				if err := sc.handleDomainEvent(event); err != nil {
-					// This isn't a case for failover, it's a case for disconnect and request a
-					// fresh hydration. Coming in next chunk of work
+					go sc.restartStreamFn()
 				}
 			}
-		case <-sc.ctx.Done():
+		case <-streamCtx.Done():
```
sighphyre, line 183: Run stream now binds to a scoped context. If that context dies the reader dies. Before this was struct level, meaning this could only be done once

streaming_fetcher.go
```diff
@@ -199,4 +230,10 @@ func (sc *streamingFetcher) snapshot() *FeatureMemoryState {
 
 func (sc *streamingFetcher) stop() {
 	sc.cancel()
+
+	sc.streamMu.Lock()
+	if sc.streamCancel != nil {
+		sc.streamCancel()
+	}
+	sc.streamMu.Unlock()
```
gastonfournier: Seems to be the common practice. It should be fine, but I'm thinking... what if streamCancel() fails... will it remain locked?

```suggestion
	sc.streamMu.Lock()
	defer sc.streamMu.Unlock()
	
	if sc.streamCancel != nil {
		sc.streamCancel()
	}
```

gastonfournier (approved):
Just one relatively minor comment (it's an edge case)

streaming_fetcher.go
sighphyre: No exceptions in Go so the unlock will still execute. Don't think cancellations can fail anyway. Still, I like the suggestion, I think it fits the style of the SDK right now and it's defensive against future mucking about in this function

PR #263, opened 2026-04-07
Feat/ready event on cutover

sighphyre:
Improves the adaptive fetcher cutover flow. The SDK now only cuts over to the next fetcher once that fetcher has signaled that its ready - in practice that it's been able to hydrate from Unleash

### DIFF
--- a/adaptive_fetcher.go
+++ b/adaptive_fetcher.go
@@ -32,7 +32,7 @@ func defaultFetcherFactory() fetcherFactory {
 
 type adaptiveFetcher struct {
 	currentFetcher        atomic.Pointer[fetchContainer]
-	pendingFetcher        togglerFetcher
+	cutoverFetcher        togglerFetcher
 	baseOptions           fetcherOptions
 	baseChannels          fetcherChannels
 	internalReady         chan bool
@@ -118,10 +118,10 @@ func (af *adaptiveFetcher) handleReadyEvent() {
 		return
 	}
 
-	if af.pendingFetcher != nil {
+	if af.cutoverFetcher != nil {
 		af.currentFetcher.Load().f.stop()
-		af.currentFetcher.Store(&fetchContainer{f: af.pendingFetcher})
-		af.pendingFetcher = nil
+		af.currentFetcher.Store(&fetchContainer{f: af.cutoverFetcher})
+		af.cutoverFetcher = nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


adaptive_fetcher.go
```diff
@@ -105,6 +104,27 @@ func (af *adaptiveFetcher) superviseFetchers() {
 	}
 }
 
+func (af *adaptiveFetcher) handleReadyEvent() {
```
sighphyre, line 107: Pulled this out into a function so we can use defer on the mutex unlock because I made myself a nice little deadlock during dev and I'd rather not do that in prod

adaptive_fetcher.go
```diff
@@ -118,16 +138,19 @@ func (af *adaptiveFetcher) cutover() {
 		return
 	}
 
+	if af.pendingFetcher != nil {
```
sighphyre: Just a sanity check, if we pending fetcher is waiting to take over we probably don't want a second one

adaptive_fetcher_test.go
```diff
@@ -251,3 +292,38 @@ func TestAdaptiveFetcher_FailoverSignalCutsOverAndWarns(t *testing.T) {
 		t.Fatalf("expected warning to be emitted on failover")
 	}
 }
+
+func TestAdaptiveFetcher_CutoverWaitsForPollingReady(t *testing.T) {
```
sighphyre, line 296: Don't like changing stuff without adding new tests but this might already be meaningfully covered by existing tests. Open to yeeting it

adaptive_fetcher.go
```diff
@@ -32,6 +32,7 @@ func defaultFetcherFactory() fetcherFactory {
 
 type adaptiveFetcher struct {
 	currentFetcher        atomic.Pointer[fetchContainer]
+	pendingFetcher        togglerFetcher
```
gastonfournier: Naming... should it be called replacementFetcher, or cutoverFetcher? I feel that pending fetcher suggests there's something pending in normal operations, when actually this is only pending when there's a cutover... otherwise it's always nil. 

Consider it a minor, I can also see reasons why it's called pendingFetcher... so maybe a short comment on the variable definition explaining the purpose would help future devs

adaptive_fetcher.go
gastonfournier: Here, cutoverFetcher would make more sense (inside the cutoverFunction) as it clearly signifies that the cutover was already set.

gastonfournier (approved):
LGTM!

adaptive_fetcher.go
sighphyre: Nah, if you have to put comments explaining variable names then something has gone terribly wrong. I can live with cutoverFetcher if you think that's clearer. I'll rename it

PR #264, opened 2026-04-08
fix: broken events now correctly fail instead of silently continuing

sighphyre:
This forces events that are corrupted to correctly fail in the delta parse code. Corrupted means one of the following is true:

- The SSE envelope is valid but the content is not valid JSON
- The SSE envelope is valid and the outer contents are valid JSON but the inner packet of the event data is not valid JSON
- The SSE envelope is valid and the content is JSON but the packet cannot be parsed as the type that `Type` specifies. This only applies if that type is one of the 5 known event types

This does not reject well formed events that are unrecognized - this gives us a layer of forwards compatibility if we decide to add new events in a backwards compatible way

Does a bunch of clean ups to improve the flow and move us away from the TypeScript in Go patterns that we were previously using

### DIFF
--- a/api/delta.go
+++ b/api/delta.go
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
 	Events []DeltaEvent `json:"events"`
 }
 
-type DeltaEvent interface {
-	GetType() string
-	GetEventId() int
+type DeltaState struct {
+	Features map[string]*Feature
+	Segments map[int][]Constraint
 }
+
+type DeltaEvent interface {
+	Apply(state *DeltaState)
+}
+
+type BaseDeltaEvent struct {
+	Type    string `json:"type"`
+	EventId int    `json:"eventId"`
+}
+
 type FeatureUpdatedEvent struct {
-	Type    string  `json:"type"`
-	EventId int     `json:"eventId"`
+	BaseDeltaEvent
 	Feature Feature `json:"feature"`
 }
 
-func (e *FeatureUpdatedEvent) GetType() string { return e.Type }
-func (e *FeatureUpdatedEvent) GetEventId() int { return e.EventId }
-
 type FeatureRemovedEvent struct {
-	Type        string `json:"type"`
-	EventId     int    `json:"eventId"`
+	BaseDeltaEvent
 	FeatureName string `json:"featureName"`
 	Project     string `json:"project"`
 }
-
-func (e *FeatureRemovedEvent) GetType() string { return e.Type }
-func (e *FeatureRemovedEvent) GetEventId() int { return e.EventId }
-
 type SegmentUpdatedEvent struct {
-	Type    string  `json:"type"`
-	EventId int     `json:"eventId"`
+	BaseDeltaEvent
 	Segment Segment `json:"segment"`
 }
 
-func (e *SegmentUpdatedEvent) GetType() string { return e.Type }
-func (e *SegmentUpdatedEvent) GetEventId() int { return e.EventId }
-
 type SegmentRemovedEvent struct {
-	Type      string `json:"type"`
-	EventId   int    `json:"eventId"`
-	SegmentId int    `json:"segmentId"`
+	BaseDeltaEvent
+	SegmentId int `json:"segmentId"`
 }
 
-func (e *SegmentRemovedEvent) GetType() string { return e.Type }
-func (e *SegmentRemovedEvent) GetEventId() int { return e.EventId }
-
 type HydrationEvent struct {
-	Type     string    `json:"type"`
-	EventId  int       `json:"eventId"`
+	BaseDeltaEvent
 	Features []Feature `json:"features"`
 	Segments []Segment `json:"segments"`
 }
 
-func (e *HydrationEvent) GetType() string { return e.Type }
-func (e *HydrationEvent) GetEventId() int { return e.EventId }
+func (e *FeatureUpdatedEvent) Apply(state *DeltaState) {
+	state.Features[e.Feature.Name] = &e.Feature
+}
 
-// UnmarshalJSON implements custom unmarshaling for ClientFeaturesDelta
+func (e *FeatureRemovedEvent) Apply(state *DeltaState) {
+	delete(state.Features, e.FeatureName)
+}
+
+func (e *SegmentUpdatedEvent) Apply(state *DeltaState) {
+	state.Segments[e.Segment.Id] = e.Segment.Constraints
+}
+
+func (e *SegmentRemovedEvent) Apply(state *DeltaState) {
+	delete(state.Segments, e.SegmentId)
+}
+
+func (e *HydrationEvent) Apply(state *DeltaState) {
+	state.Features = make(map[string]*Feature, len(e.Features))
+	for _, f := range e.Features {
+		state.Features[f.Name] = &f
+	}
+
+	state.Segments = make(map[int][]Constraint, len(e.Segments))
+	for _, seg := range e.Segments {
+		state.Segments[seg.Id] = seg.Constraints
+	}
+}
+
+func unmarshalEvent[T DeltaEvent](raw json.RawMessage) (DeltaEvent, error) {
+	var event T
+	if err := json.Unmarshal(raw, &event); err != nil {
+		return nil, err
+	}
+	return event, nil
+}
+
 func (c *ClientFeaturesDelta) UnmarshalJSON(data []byte) error {
-	// First unmarshal to get the raw events
 	var raw struct {
 		Events []json.RawMessage `json:"events"`
 	}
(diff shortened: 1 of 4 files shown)


api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
 }
 
+type DeltaState struct {
+	Features map[string]*Feature
+	Segments map[int][]Constraint
+}
+
 type DeltaEvent interface {
-	GetType() string
-	GetEventId() int
```
sighphyre, line 14: This is a poor way to model these events. Because Go is structurally typed, it means we can pass anything that implements these two methods. What's notably missing here is... well... everything that makes an event important, so this interface is actively harmful

api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
+	Apply(state *DeltaState)
```
sighphyre, line 18: Now an event is anything that can be applied to a delta state - basically a wrapper for features + segments. This gives us two nice things:
1) We no longer have to switch on their type as we parse as well as when we apply them - that's a pretty fragile approach which splits their meaning into two different domains
2) Obeys open/closed, which means we don't have to muck with the delta processor when we want to make changes to what events we ingest

api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
 }
+
+type BaseDeltaEvent struct {
+	Type    string `json:"type"`
+	EventId int    `json:"eventId"`
+}
```
sighphyre, line 24: Honestly just more idiomatic Go to use struct embedding than redefining these fields 5 times

api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
-	Type     string    `json:"type"`
-	EventId  int       `json:"eventId"`
+	BaseDeltaEvent
 	Features []Feature `json:"features"`
 	Segments []Segment `json:"segments"`
 }
 
-func (e *HydrationEvent) GetType() string { return e.Type }
-func (e *HydrationEvent) GetEventId() int { return e.EventId }
+func (e *FeatureUpdatedEvent) Apply(state *DeltaState) {
```
sighphyre, line 52: Lifted from delta processor - events know how to apply a patch to a feature state to give us before -> after

api/delta.go
```diff
@@ -71,69 +94,44 @@ func (c *ClientFeaturesDelta) UnmarshalJSON(data []byte) error {
...
 
 	for _, rawEvent := range raw.Events {
-		// Determine the event type
 		var eventType struct {
 			Type string `json:"type"`
 		}
 
 		if err := json.Unmarshal(rawEvent, &eventType); err != nil {
-			continue // Skip malformed events
+			return fmt.Errorf("failed to unmarshal event type: %w", err)
```
sighphyre, line 105: Intentional change - if our JSON failed to parse we know a few things - we have a healthy SSE envelope but that envelope contains borked JSON. It's not safe to continue in this case - Unleash has attempted to update something but it's broken. Raising an error here drives us through the disconnect + rehydrate flow

api/delta.go
```diff
@@ -71,69 +94,44 @@ func (c *ClientFeaturesDelta) UnmarshalJSON(data []byte) error {
...
+			event, err = unmarshalEvent[*HydrationEvent](rawEvent)
 		default:
-			// Unknown event type - skip
+			// Unknown event type, this gives us a safety net for forward compatibility with new event types
 			continue
 		}
 
-		if event != nil {
-			c.Events = append(c.Events, event)
+		if err != nil {
```
sighphyre, line 129: Another intentional change - we do not skip broken events. If we know that we expect a feature updated event and we cannot parse than event, it's not safe to continue. Our stream is corrupted and we need to attempt a reconnect

api/delta.go
```diff
@@ -71,69 +94,44 @@ func (c *ClientFeaturesDelta) UnmarshalJSON(data []byte) error {
...
+			return fmt.Errorf("failed to unmarshal %s event: %w", eventType.Type, err)
 		}
+
+		c.Events = append(c.Events, event)
 	}
 
 	return nil
 }
-
-func ParseDelta(data []byte) (*ClientFeaturesDelta, error) {
```
sighphyre, line 133: This isn't helpful, it just wraps Unmarshall. Callers are capable of doing that themselves without wrapping this. Open to bringing it back if anyone wants but I think it's just noise

api/delta_test.go
```diff
@@ -2,367 +2,9 @@ package api
 
 import (
 	"encoding/json"
-	"reflect"
 	"testing"
 )
 
-func TestParseDeltaEvent(t *testing.T) {
```
sighphyre, line 9: Most of the tests already parse. Not useful. Yeeted

api/delta_test.go
```diff
@@ -2,367 +2,9 @@ package api
...
-			}
-
-			if tt.wantCount > 0 && delta.Events[0].GetType() != tt.wantType {
-				t.Errorf("First event type = %v, want %v", delta.Events[0].GetType(), tt.wantType)
-			}
-		})
-	}
-}
-
-func TestDeltaEventTypes(t *testing.T) {
```
sighphyre, line 171: Types don't actually matter. What matters is that they're applied correctly. Also yeeted

api/delta_test.go
```diff
@@ -381,117 +23,13 @@ func TestEventOrdering(t *testing.T) {
 	// Events should maintain their order as received
 	expectedIds := []int{3, 1, 2}
 	for i, event := range delta.Events {
-		if event.GetEventId() != expectedIds[i] {
-			t.Errorf("Event %d has ID %d, want %d", i, event.GetEventId(), expectedIds[i])
-		}
-	}
-}
-
-func TestDeltaEventInterface(t *testing.T) {
```
sighphyre, line 390: Also not useful. Interfaces are not a domain constraint

api/delta_test.go
```diff
@@ -381,117 +23,13 @@ func TestEventOrdering(t *testing.T) {
...
-				t.Errorf("GetType() = %v, want %v", got, tt.wantType)
-			}
-			if got := tt.event.GetEventId(); got != tt.wantId {
-				t.Errorf("GetEventId() = %v, want %v", got, tt.wantId)
-			}
-		})
-	}
-}
-
-func TestUnmarshalDeltaEvents(t *testing.T) {
```
sighphyre, line 441: Another check to see if we can parse. Not helpful. Yeeted

api/delta_test.go
```diff
@@ -381,117 +23,13 @@ func TestEventOrdering(t *testing.T) {
...
 		}
-	} else {
-		t.Error("First event should be FeatureUpdatedEvent")
-	}
-
-	if fr, ok := delta.Events[1].(*FeatureRemovedEvent); ok {
-		if fr.FeatureName != "f2" {
-			t.Errorf("FeatureRemoved feature name = %v, want f2", fr.FeatureName)
+		if id != expectedIds[i] {
+			t.Errorf("Event %d has ID %d, want %d", i, id, expectedIds[i])
```
sighphyre, line 31: Github is being unhelpful here - this is part of the previous test, not a mutation to what we see on the left

api/delta_test.go
```diff
@@ -538,32 +79,113 @@ func TestDeltaWithUnknownEventType(t *testing.T) {
...
+	case *SegmentRemovedEvent:
+		return e.EventId, true
+	case *HydrationEvent:
+		return e.EventId, true
+	default:
+		return 0, false
 	}
+}
+
+func TestApplyFeatureUpdatedEvent(t *testing.T) {
```
sighphyre, line 99: New tests to cover our Apply logic. This is probably the kind of stuff I actually want covered by integration tests but I don't think it's harmful to have these in unit test form too

delta_processor.go
```diff
@@ -59,41 +59,18 @@ func (dp *deltaProcessor) process(delta *api.ClientFeaturesDelta) (*api.FeatureR
...
-			featMap = make(map[string]*api.Feature, len(e.Features))
-			for _, f := range e.Features {
-				featMap[f.Name] = &f
-			}
-
-			segMap = make(map[int][]api.Constraint, len(e.Segments))
-			for _, seg := range e.Segments {
-				segMap[seg.Id] = seg.Constraints
-			}
+	deltaState := &api.DeltaState{
```
sighphyre, line 62: deltaProcessor no longer needs to know how to apply the events, only that an event is applicable to the current state

This is kinda important from an error handling perspective because now borked events are handled in one and only one place - the parse. The default case in this switch statement wasn't preventing weird events from Unleash landing here, it was silently masking programmer error where someone could pass something that conformed to the interface but wasn't an event.

api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
-	EventId  int       `json:"eventId"`
+	BaseDeltaEvent
 	Features []Feature `json:"features"`
 	Segments []Segment `json:"segments"`
 }
 
-func (e *HydrationEvent) GetType() string { return e.Type }
-func (e *HydrationEvent) GetEventId() int { return e.EventId }
+func (e *FeatureUpdatedEvent) Apply(state *DeltaState) {
+	state.Features[e.Feature.Name] = &e.Feature
```
gastonfournier, line 53: What happens if the feature name is new? i.e. a new feature added? does the size of the state.Features need to be adapted?

api/delta.go
```diff
@@ -9,60 +9,83 @@ type ClientFeaturesDelta struct {
...
+}
+
+func (e *FeatureRemovedEvent) Apply(state *DeltaState) {
+	delete(state.Features, e.FeatureName)
+}
+
+func (e *SegmentUpdatedEvent) Apply(state *DeltaState) {
+	state.Segments[e.Segment.Id] = e.Segment.Constraints
```
gastonfournier, line 61: Same with segments, a new used segment will appear as a segment updated. If this appends to the last and increases the size of the array, we're good

api/delta.go
gastonfournier: Should we return a value as an error? Should we have something like return nil, fmt.Error?

Not sure if this is fine when you have nothing to return, and therefore the only thing left is the error, if present.

delta_processor.go
gastonfournier: Just to confirm: is this a copy of featMap and featMap? because we're applying patches on the fly and then swapping.

gastonfournier (approved):
LGTM!

api/delta.go
sighphyre: Size does matter, yeah. Because this is Go and it's deeply committed to being confusing - the size is automatically updated

api/delta.go
sighphyre: Yep it does

delta_processor.go
sighphyre: Yep. Moving the values into DeltaState here is a copy. From here on we mutate them in place

api/delta.go
sighphyre: Nah, this is pretty normal. Adding an always nil return value here just makes life confusing for the caller

PR #265, opened 2026-04-09
feat: make delta api work

sighphyre:
This is actually primarily a patch to make the client specification tests work against production code for streaming/delta events - the spec tests for delta are intentionally built so that you can't make them pass without correctly implementing delta API. This patch makes delta API work and removes the hacks that were allowing the spec tests to pass.

We now accept a union type on the API of either a delta or a full response and then just hand that off to delta processor to make a decision around either applying the full update or layering in the events.

Delta processor now needs a rename, done in #266

### DIFF
--- a/spec_test.go
+++ b/spec_test.go
@@ -25,13 +25,6 @@ const specFolder = "./testdata/client-specification/specifications"
 var specIndex = filepath.Join(specFolder, "index.json")
 var specNotImplemented = []string{""}
 
-type TestState struct {
-	Version  int               `json:"version"`
-	Features []api.Feature     `json:"features"`
-	Segments []api.Segment     `json:"segments"`
-	Events   []json.RawMessage `json:"events"`
-}
-
 type TestCase struct {
 	Description    string          `json:"description"`
 	Context        context.Context `json:"context"`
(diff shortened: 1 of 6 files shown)


spec_test.go
```diff
@@ -106,34 +99,21 @@ func (vtc VariantTestCase) RunWithClient(client *Client) func(*testing.T) {
 
 type TestDefinition struct {
 	Name         string            `json:"name"`
-	State        TestState         `json:"state"`
+	State        json.RawMessage   `json:"state"`
```
sighphyre, line 102: There's no reason to be parsing this, makes this suite fragile. We accept what's there and pass it to the API. If the API can't handle it then the API is broken and this should fail

spec_test.go
```diff
@@ -142,86 +122,6 @@ func (td TestDefinition) Mock(listener interface{}) (*Client, error) {
 	)
 }
 
-func (td TestDefinition) processDeltaEvents() ([]api.Feature, []api.Segment) {
```
sighphyre, line 145: ☹️ ☹️ ☹️

chriswk (approved):
I like the lines delta here. Also understand the change for the processor. have a 👍

PR #266, opened 2026-04-09
chore: renamme delta processor to feature cache

sighphyre:
100% rename refactor

Bit of a more honest name now. Feature cache closes over a cache of features and allows controlled updates and quick lock free reads.


chriswk (approved):
👍

PR #270, opened 2026-07-01
fix: flush pending metrics on Close

### LINKED ISSUE #269: Client.Close() drops metrics recorded since the last periodic flush
### Summary

When `Client.Close()` is called, any toggle counts and impact metrics accumulated between the last periodic tick and shutdown are silently dropped. There is no way for the caller to request a graceful shutdown that flushes buffered metrics.

### Root cause

`metrics.Close()` (in `metrics.go`) stops the ticker, cancels the request context, and signals the sync loop to exit. The sync loop's shutdown case just returns — it never calls `sendMetrics` on the buffered counters.

```go
case <-m.close:
    close(m.closed)
    return
```

### Impact

In workloads with a metrics interval of a minute (default) and short-lived flag evaluations right before shutdown, up to a full interval's worth of counts is lost on every clean exit — including `defer client.Close()` in short-running jobs, CLIs, and any process that exits soon after evaluating flags.

### Proposed fix

Add a `Client.Shutdown(ctx context.Context) error` method mirroring `net/http.Server.Shutdown`. `Close()` stays as the immediate/fast teardown; `Shutdown(ctx)` performs a best-effort final flush bounded by the caller's context. PR incoming.

AliMsayleb:
## Summary
`Client.Close()` today drops any toggle counts and impact metrics accumulated between the last periodic tick and shutdown. This PR makes `Close()` do a best-effort final flush before returning, bounded internally so an unreachable metrics endpoint can't hang shutdown.

The flush is deliberately silent (no `OnSent`/`OnError`, no backoff-counter change) because the caller has committed to teardown and cannot react to those events.

`Close()`'s signature is unchanged — this is a pure behavior fix, no API surface change.

Fixes #269.

## Notes on prior revision
An earlier version of this PR added a separate `Shutdown(ctx)` method inspired by `net/http.Server`. Per @sighphyre's review, Unleash SDKs standardize on a single lifecycle method across the ecosystem, so that was folded back into `Close()`. The bound is a hardcoded 5s; if there's later demand for tuning it, an option can be added in a follow-up.

## Test plan
- [x] `TestMetrics_CloseFlushesBufferedMetrics` — with a 1-hour tick interval so no periodic flush can have run, asserts that `Close()` still POSTs the buffered counts to `/client/metrics`
- [x] `make` (vet + tests) green
- [x] `make test-race` green

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,7 +1,6 @@
 package unleash
 
 import (
-	stdcontext "context"
 	"fmt"
 
 	"net/http"
@@ -455,9 +454,10 @@ func (uc *Client) ImpactMetrics() *impactmetrics.MetricsAPI {
 	return uc.MetricsAPI
 }
 
-// Close stops the client from syncing data from the server. Any metrics
-// recorded since the last periodic flush are discarded. For a graceful
-// shutdown that attempts to flush buffered metrics, use Shutdown.
+// Close stops the client from syncing data from the server. It makes a
+// best-effort attempt to POST any metrics buffered since the last periodic
+// flush before returning; that final flush is bounded internally so an
+// unreachable metrics endpoint cannot hang shutdown indefinitely.
 func (uc *Client) Close() error {
 	uc.fetcher.stop()
 	uc.metrics.Close()
@@ -467,21 +467,6 @@ func (uc *Client) Close() error {
 		<-uc.closed
 	}
 	return nil
-}
-
-// Shutdown gracefully stops the client. It performs the same teardown as
-// Close but first makes a best-effort attempt to POST any metrics buffered
-// since the last periodic flush. The provided context bounds how long the
-// flush is allowed to take; if it expires, remaining metrics are dropped
-// and Shutdown returns ctx.Err().
-func (uc *Client) Shutdown(ctx stdcontext.Context) error {
-	uc.fetcher.stop()
-	err := uc.metrics.Shutdown(ctx)
-	if uc.options.listener != nil {
-		close(uc.close)
-		<-uc.closed
-	}
-	return err
 }
 
 // Errors returns the error channel for the client.
--- a/client.go
+++ b/client.go
@@ -1,6 +1,7 @@
 package unleash
 
 import (
+	stdcontext "context"
 	"fmt"
 
 	"net/http"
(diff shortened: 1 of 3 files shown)


sighphyre:
Hey @AliMsayleb, thanks for taking a stab here. I'm open to accepting a fix here but honestly I think `Close` not sending metrics is just a missing feature, so I'd really rather we just make Close do the send here and not add a second method. 

If you're willing to patch this to do that I'm happy to give it a more thorough review and work towards merging it. Shout if you don't have time and we'll get to it when we have time on our roadmap

AliMsayleb:
> Hey @AliMsayleb, thanks for taking a stab here. I'm open to accepting a fix here but honestly I think `Close` not sending metrics is just a missing feature, so I'd really rather we just make Close do the send here and not add a second method.
> 
> If you're willing to patch this to do that I'm happy to give it a more thorough review and work towards merging it. Shout if you don't have time and we'll get to it when we have time on our roadmap

If you check `net/http.Server` it provides both `Close` and `Shutdown`
And this is the `Close` doc:
```
// Close immediately closes all active net.Listeners and any
// connections in state [StateNew], [StateActive], or [StateIdle]. For a
// graceful shutdown, use [Server.Shutdown].
```

I used http.Server as an inspiration to add a new method rather than making `Close` blocking until it syncs the last remaining metrics. 
What do you think @sighphyre ?

sighphyre:
@AliMsayleb Yeah fair I can see why you'd want this based off the std lib. That being said, this API surface behavior on SDKs needs to standardize well across all of our SDKs. So it's often less informed by what about what Go chooses to do and more about how our ecosystem is intended to work - one lifecycle method that shuts the client down and flushes pending metrics as part of teardown.

I’m also more comfortable with Close() becoming slightly blocking than I am with users in the wild silently losing metrics unless they discover a new method. Losing the final metrics batch feels like the bug here, so I’d rather fix that behavior directly in Close() instead of adding a second API path.

So my preference is still: no new Shutdown(ctx) method for now; make Close() do a best-effort bounded flush before closing.

AliMsayleb:
@sighphyre thanks for the feedback! I went with your suggested approach: dropped the separate `Shutdown(ctx)` method and folded the best-effort flush into `Close()` itself with a 5s internal bound. `Close()`'s signature is unchanged — pure behavior fix.

Ready for another look when you have a moment 🙏

sighphyre (approved):
Awesome, looks pretty reasonable to me and I don't see anything that would be dangerous here

sighphyre:
Thanks again @AliMsayleb, released in 6.5.0

